PR #25, opened 2015-07-28
Add configurable connection and read timeouts

llehtinen:
I was toying with a Bugsnag logback appender which was calling client.notify inside the synchronized append method. Thread dump showed one thread stuck in `int status = conn.getResponseCode();` inside HttpClient, and a few hundred threads waiting for it to release the lock it was holding.

Apparently if not configured, the connect and read timeouts are infinite. I added 60s defaults and tried to make them configurable. Let me know what you think.


kattrali:
Thanks, @llehtinen. This looks good. Merging after I add some documentation around the additional properties.

Issue #23, opened 2015-03-06
Send hostname


snmaynard:
We should send the hostname on java

kattrali:
Fixed by #30

PR #30, opened 2016-01-14
Add hostname

bullmo:
See https://github.com/bugsnag/bugsnag-java/issues/23


bullmo:
There is an issue with using InetAddress.getLocalHost() on openJDK (see travis failure). I will try to implement an alternative approach.

bullmo:
Get the hostname (or computer name) from the environment (if it is available).

kattrali:
Looks good! :+1:

PR #34, opened 2016-02-23
Use SSL by default for communications with Bugsnag.com

duncanhewett:
Changes the default to SSL in line with other notifiers and highlights the ability to change it in the documentation.


kattrali:
Looks good. Thanks again, @duncanhewett!

PR #35, opened 2016-02-27
Truncate very long payloads

duncanhewett:
Truncate very long payloads which will get rejected by Bugsnag with a 400 response.

Long values from the payload are limited in length and the string "[TRUNCATED]" is appended.  This is the same strategy used by the Bugsnag Ruby notifier - see:
https://github.com/bugsnag/bugsnag-ruby/blob/eb782074fa127bb6ede77f31347672f65591489c/lib/bugsnag/helpers.rb#L16


kattrali:
Thanks, @duncanhewett!

PR #36, opened 2016-03-23
Added a thread pool to allow asynchronous notification to Bugsnag

bullmo:
Adds a thread pool with 2 daemon worker threads that allow the notification to Bugsnag to occur asynchronously. Notifications are queued if the worker threads are busy.

Async notification is turned on by default but synchronous notification can be used using the `setAsynchronousNotification` method

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,16 @@ Set whether communication with Bugsnag.com should be made via SSL. To use HTTP s
 bugsnag.setUseSSL(false);
 ```
 
+###setAsynchronousNotification
+
+Sets whether error notifications to Bugsnag.com are sent asynchronously.
+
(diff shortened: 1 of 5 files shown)


kattrali:
Looks good, thanks @bullmo. Async will be the new default, which will warrant a major bump, but I think that's fine. What do you think, @snmaynard?

snmaynard:
Is there a limit on the number of queued notifications?

I'm ok with this going out as a minor bump tbh as its forward compat - unless im missing something?

kattrali:
It breaks for anyone depending on synchronous upload behavior ¯\_(ツ)_/¯

snmaynard:
True - your call :D

kattrali:
:shipit:

PR #40, opened 2016-05-02
fix contributing link

ndurell:
Link appeared to be broken but I'm not sure if I pointed it to the right place.


kattrali:
Thanks @ndurell!

PR #46, opened 2016-08-22
[WIP] Added a slf4j logger

GrahamCampbell:


pom.xml
```diff
@@ -47,6 +47,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.slf4j</groupId>
```
GrahamCampbell: Do we need to add this as a hard dependency? Does maven have a way to "suggest" dependencies?

src/main/java/com/bugsnag/BugsnagLogger.java
```diff
@@ -0,0 +1,263 @@
...
+    public String getName() {
+        return LOGGER_NAME;
+    }
+
+    public boolean isTraceEnabled() {
+        return false;
+    }
+
+    public void trace(String msg) {
+        // Trace not supported
```
GrahamCampbell: Once we have breadcrumbs, we can implement trace to leave a breadcrumb. Same for info.

src/main/java/com/bugsnag/BugsnagLogger.java
```diff
@@ -0,0 +1,263 @@
...
+    public void debug(Marker marker, String msg, Throwable t) {
+        // Markers not supported
+    }
+
+    public boolean isInfoEnabled() {
+        return true;
+    }
+
+    public void info(String msg) {
+        // client.notify(msg, "info");
```
GrahamCampbell: Strings not supported yet. Probably will end up in a separate PR.

GrahamCampbell:
TODO: Add tests.

pom.xml
foygl: I think this is fine as it's just an API and is not bringing in any heavyweight loggers. SLF4J have their own opinion on this: http://slf4j.org/faq.html#optional_dependency

src/main/java/com/bugsnag/BugsnagLogger.java
```diff
@@ -0,0 +1,263 @@
+package com.bugsnag;
+
+import org.slf4j.Logger;
+import org.slf4j.Marker;
+import org.slf4j.helpers.MessageFormatter;
+
+public class BugsnagLogger implements Logger {
```
foygl: Probably good to have a class level javadoc about what this does.

foygl:
It feels like for this logger to be useful, it needs to still log to the console as well as notifying Bugsnag. Not sure of the best way to do this without bringing in another dependency - we want to avoid bringing in anything like log4j as a dependency if we can help it.

foygl:
As a further thought, we possibly want to implement appenders for logback-classic, log4j and java.util.logging which all speak SLF4J, and possibly Jakarta (JCL) although I'm not sure how this plays with SLF4J. Then people can just use the appender appropriate to their current logging framework alongside their default appender.

logback-classic:
 - http://logback.qos.ch/manual/configuration.html#configuringAppenders
 - http://logback.qos.ch/apidocs/ch/qos/logback/core/Appender.html

log4j:
 - https://logging.apache.org/log4j/2.x/manual/appenders.html
 - https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/appender/AbstractAppender.html

java.util.logging:
 - https://docs.oracle.com/javase/8/docs/api/java/util/logging/Handler.html
- https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html#addHandler-java.util.logging.Handler-

If we do this, we need to consider the implications of bringing in even more dependencies. If we need anything more than APIs we should probably split out into modules for each appender...

kattrali:
> It feels like for this logger to be useful, it needs to still log to the console as well as notifying Bugsnag. 

Agreed. In the past when I’ve had multiple destinations, I ended up constructing factory with more than one logger name, and then generated logger instances had other loggers under the hood. It sounds more complex than it was as in the end it was maybe 30 lines of code. We could package a similar factory class though it may or may not be necessary.

> As a further thought, we possibly want to implement appenders for logback-classic, log4j and java.util.logging which all speak SLF4J, and possibly Jakarta (JCL) although I'm not sure how this plays with SLF4J. Then people can just use the appender appropriate to their current logging framework alongside their default appender.

Agreed. On the dependency front it would make the most sense for each dependency to get its own library, (e.g. “bugsnag-java-log4j”) as logger dependencies from libraries are usually the worst kind. The appender interfaces are fairly straightforward though.

src/main/java/com/bugsnag/BugsnagLogger.java
GrahamCampbell: Just discussed this. Bad idea basically. Will not be supported ever. :P

kattrali:
The interface here got a bit shuffled around while shipping v3, but logging support is now baked in! :D

PR #47, opened 2016-08-25
New major version of notifier

loopj:
## ChangeLog
-   Uses [Jackson](https://github.com/FasterXML/jackson-databind) for streaming JSON serialization
-   Uses [Gradle](https://gradle.org) as the build system
-   Logs internally using `org.slf4j.Logger`
-   `Severity` is now an enum instead of a `String`
-   Request information is automatically collected in Servlet API apps
-   JVM runtime, O/S, and locale diagnostics are collected
-   The `Client` object has been renamed to `Bugsnag`
-   The `Event` object has been renamed to `Report`
-   `Report` object is now exposed for ease of attaching diagnostics to error reports
-   Targets Java 1.6
-   Callbacks now support Java 8 lambda syntax, eg:
  
  ``` java
  bugsnag.addCallback((report) -> {
      report.setSeverity(Severity.ERROR);
  });
  ```
-   Chaining support added to `Report` methods
  
  ``` java
  bugsnag.addCallback((report) -> {
      report.setSeverity(Severity.ERROR).setUserId("123");
  });
  ```
-   Can now add a Callback to `Bugsnag.notify` calls
  
  ``` java
  bugsnag.notify(ex, (report) -> {
      report.addToTab("tab", "key", "value");
  });
  ```
-   Can now change the API key on a per-report basis using Callbacks
  
  ``` java
  bugsnag.notify(ex, (report) -> {
      report.setApiKey("123456");
  });
  ```
-   Error report delivery is now fully swappable via the `Delivery` interface
  
  ``` java
  // Low-level use
  Delivery delivery = new AsyncHttpDelivery();
  delivery.setEndpoint(...);
  bugsnag.setDelivery(delivery);
  
  // Use synchronous delivery instead
  Delivery delivery = new SyncHttpDelivery();
  bugsnag.setDelivery(delivery);
  
  // Shortcuts
  bugsnag.setEndpoint(...);
  bugsnag.setProxy(...);
  bugsnag.setTimeout(...);
  ```


kattrali:
> 1 commit

> +1,739 −1,808 changed lines

… challenge accepted.

kattrali:
FYI a couple of `*.class` files were checked in

kattrali:
And build files within `examples/build`

src/main/java/com/bugsnag/Client.java
```diff
@@ -1,208 +1,357 @@
...
-    public void addToTab(String tab, String key, Object value) {
-        config.addToTab(tab, key, value);
-    }
 
-    public void clearTab(String tab) {
-        config.clearTab(tab);
-    }
+        // Don't notify if this error class should be ignored
+        if (config.shouldIgnoreClass(report.getExceptionName())) {
+            logger.debug("Error not reported to Bugsnag - exception class is in 'ignoreClasses'");
```
kattrali: Probably should log the class name in this message, for clarity

src/main/java/com/bugsnag/Client.java
```diff
@@ -1,208 +1,357 @@
...
 
-    public void trackUser() {
-        try {
-            Metrics metrics = new Metrics(config, diagnostics);
-            metrics.deliver();
-        } catch (NetworkException ex) {
-            config.logger.warn("Error sending metrics to Bugsnag", ex);
+        // Don't notify unless releaseStage is in notifyReleaseStages
+        if (!config.shouldNotifyForReleaseStage()) {
+            logger.debug("Error not reported to Bugsnag - 'releaseStage' is not in 'notifyReleaseStages'");
```
kattrali: The name of the release stage is not in the message

src/main/java/com/bugsnag/delivery/OutputStreamDelivery.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public OutputStreamDelivery(OutputStream outputStream) {
+        this.outputStream = outputStream;
+    }
+
+    public void deliver(Serializer serializer, Object object) {
+        try {
+            serializer.writeToStream(outputStream, object);
+        } catch (SerializationException ex) {
+            // Meh
```
kattrali: probably should at least log this

duncanhewett:
@loopj It was just a best practice/style decision, but I've reverted that commit.

duncanhewett:
To get the build to work in all the JDKs I have:
- Switched to Gradle 2.14 for JDK6 support
- Changed the Travis build to use `jdk_switcher` to run different builds per JDK
- Only build example apps on JDK8 (so they can use Lambda syntax)
- Only run checkstyle on JDK8 (plugin doesn't support JDK6)
- Use jackson-databind version 2.6.7 for JDK6 compability

src/main/java/com/bugsnag/Report.java
```diff
@@ -0,0 +1,202 @@
...
+        return PAYLOAD_VERSION;
+    }
+
+    /**
+     * Get the exceptions for the report.
+     *
+     * @return the exceptions that make up the error.
+     */
+    @Expose
+    public List<Exception> getExceptions() {
```
foygl: This either needs to be `protected` or `Exception` needs to be public.

src/main/java/com/bugsnag/Report.java
```diff
@@ -0,0 +1,202 @@
...
+        while (currentThrowable != null) {
+            exceptions.add(new Exception(config, currentThrowable));
+            currentThrowable = currentThrowable.getCause();
+        }
+
+        return exceptions;
+    }
+
+    @Expose
+    public List<ThreadState> getThreads() {
```
foygl: This either needs to be `protected` or `ThreadState` needs to be public.

src/main/java/com/bugsnag/Report.java
```diff
@@ -0,0 +1,202 @@
...
+    private Diagnostics diagnostics = new Diagnostics();
+    private boolean shouldCancel = false;
+
+    /**
+     * Create a report for the error.
+     *
+     * @param config    the configuration for the report.
+     * @param throwable the error to create the report for.
+     */
+    public Report(Configuration config, Throwable throwable) {
```
foygl: If the expectation is that this always get built from Bugsnag.buildReport we should remove the `public` here to make this package protected

src/main/java/com/bugsnag/Report.java
duncanhewett: Changed, I don't think it needs to be public.

src/main/java/com/bugsnag/Report.java
duncanhewett: Changed, I don't think it needs to be public.

src/main/java/com/bugsnag/Bugsnag.java
```diff
@@ -0,0 +1,361 @@
...
+
+    /**
+     * Notify Bugsnag of a handled exception - with a severity.
+     *
+     * @param throwable the exception to send to Bugsnag
+     * @param severity  the severity of the error, one of {#link Severity#ERROR},
+     *                  {@link Severity#WARNING} or {@link Severity#INFO}
+     * @return true unless the error report was ignored
+     */
+    public boolean notify(Throwable throwable, Severity severity) {
```
foygl: Do we need a `notify(Throwable, Severity, Callback)` for completeness?

src/main/java/com/bugsnag/MetaData.java
```diff
@@ -1,85 +1,26 @@
...
-            JSONObject tab = getTab(tabName);
-            
-            @SuppressWarnings("unchecked")
-            Map<String, Object> mapValue = (Map<String, Object>)value;
-            
-            for (Map.Entry<String, Object> entry : mapValue.entrySet()) {
-                JSONUtils.safePut(tab, entry.getKey(), entry.getValue());
-            }
-        } else {
-            addToTab("Custom Data", tabName, value);
```
foygl: Are we officially dropping support for adding data without specifying a tab?

src/main/java/com/bugsnag/Bugsnag.java
loopj: I think we should add this yeah

loopj:
Omg

PR #48, opened 2016-11-07
[WIP] Bugsnag Logback appender

duncanhewett:
Adds a [Logback](http://logback.qos.ch/) appender to report logs containing Throwables to Bugsnag.

For example, this log call would result in a notification to Bugsnag:
`logger.error("Error will be reported to Bugsnag", new Throwable("Test error"));`

A separate module which would be added as a dependency to projects using Logback. Projects would be required to implement `BugsnagAppender.initBugsnag()` to return a `Bugsnag` to use for notifying.

TODO:
* [ ] Documentation
* [ ] Build configuration to produce a separate JAR

### DIFF
--- a/examples/logback/README.md
+++ b/examples/logback/README.md
@@ -1,0 +1,16 @@
+# Bugsnag Logback Java Example
+
+An example application to show how to use Bugsnag to report exceptions logged using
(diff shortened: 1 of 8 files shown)


fractalwrench:
Closing as this will be addressed in #90

PR #49, opened 2016-11-07
Add Spring example apps

foygl:
- Add Spring boot example app.
- Add Spring boot web based example app.
- Add `close` method to Bugsnag to ensure it tears down connections (this is automatically called by Spring when tearing down the bean).
- Change the simple example to use `close` rather than `Shutdown`.


duncanhewett:
LGTM

PR #50, opened 2016-11-07
Stop AsyncHttpDelivery from indefinitely blocking exit

foygl:
The executor used by AsyncHttpDelivery indefinitely blocks application exit unless a shutdown is manually triggered. This replaces the executor with one which times out when idle and so doesn't block an exit.

cc @bullmo


bullmo:
Looks good @foygl

PR #63, opened 2017-10-10
Allow overriding context set by servlet callback

kattrali:
Given a report generated like so and captured from a servlet context:

```java
Report report = bugsnag.buildReport(exception);
report.setContext("custom context");
bugsnag.notify(report);
```

The context in the final report should retain "custom context" and not be overridden by callbacks.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,8 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.slf4j:slf4j-jdk14:1.7.25'
+    testCompile 'javax.servlet:javax.servlet-api:3.1.0'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -32,6 +32,8 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.slf4j:slf4j-jdk14:1.7.25'
+    testCompile 'javax.servlet:javax.servlet-api:3.1.0'
+    testCompile 'org.mockito:mockito-core:2.+'
```
fractalwrench: It would be good to lock this to a specific version.

PR #69, opened 2018-01-03
Remove dependency on Guava

fractalwrench:
Removes the library's dependency on Guava, as it is not used extensively, has led to non-reproducible dependency issues, and adds quite a bit of size to built artefacts.


src/main/java/com/bugsnag/callbacks/DeviceCallback.java
```diff
@@ -2,47 +2,60 @@
...
-    private static final Supplier<String> hostnameCache =
-        Suppliers.memoize(new Supplier<String>() {
 
-            public String get() {
-                // Windows always sets COMPUTERNAME
-                if (System.getProperty("os.name").startsWith("Windows")) {
-                    return System.getenv("COMPUTERNAME");
-                }
+    private static volatile String hostname;
+    private static transient volatile boolean hostnameInitialised;
```
foygl: Any reason not to use AtomicBoolean here?

foygl: I guess it is blocking so could slow things down slightly... carry on!

fractalwrench:
Also fixes a bug in Gradle 4.0 not recognising JDK 9.0.1:  https://github.com/travis-ci/travis-ci/issues/8975

PR #77, opened 2018-03-13
Make preemptive copy of map filtering specified keys

lehphyro:
After upgrading to version 3.1.5, we noticed the following error:
```
com.bugsnag.delivery.SyncHttpDelivery    : Error not reported to Bugsnag - exception when serializing payload
com.bugsnag.serialization.SerializationException: Exception during serialization
    at com.bugsnag.serialization.Serializer.writeToStream(Serializer.java:36) ~[bugsnag-3.1.5.jar:na]
    at com.bugsnag.delivery.SyncHttpDelivery.deliver(SyncHttpDelivery.java:57) ~[bugsnag-3.1.5.jar:na]
    at com.bugsnag.delivery.AsyncHttpDelivery$2.run(AsyncHttpDelivery.java:71) [bugsnag-3.1.5.jar:na]
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_162]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_162]
    at java.lang.Thread.run(Thread.java:748) [na:1.8.0_162]
Caused by: com.fasterxml.jackson.databind.JsonMappingException: (was java.lang.UnsupportedOperationException) (through reference chain: com.bugsnag.Notification["events"]->java.util.Collections$SingletonList[0]->com.bugsnag.Report["metaData"]->com.bugsnag.util.FilteredMap["request"]->com.bugsnag.util.FilteredMap["params"])
    at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:388) ~[jackson-databind-2.8.11.1.jar:2.8.11.1]
    at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:348) ~[jackson-databind-2.8.11.1.jar:2.8.11.1]
    at com.fasterxml.jackson.databind.ser.std.StdSerializer.wrapAndThrow(StdSerializer.java:343) ~[jackson-databind-2.8.11.1.jar:2.8.11.1]
    at com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeFields(MapSerializer.java:637) ~[jackson-databind-2.8.11.1.jar:2.8.11.1]
    at com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:536) ~[jackson-databind-2.8.11.1.jar:2.8.11.1]
[...]

PR #79, opened 2018-03-21
Logback appender

lehphyro:
As promised in https://github.com/bugsnag/bugsnag-java/pull/77, here's a logback appender that supports configuration of most bugsnag client library properties and takes values from MDC for further customization of reports.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,9 +27,10 @@ repositories {
 dependencies {
     compile 'com.fasterxml.jackson.core:jackson-databind:2.9.1'
     compile 'org.slf4j:slf4j-api:1.7.25'
-    compile 'ch.qos.logback:logback-classic:1.2.3'
+    compileOnly 'ch.qos.logback:logback-classic:1.2.3'
     compileOnly 'javax.servlet:javax.servlet-api:3.1.0'
 
+    testCompile 'ch.qos.logback:logback-classic:1.2.3'
     testCompile 'junit:junit:4.12'
     testCompile 'javax.servlet:javax.servlet-api:3.1.0'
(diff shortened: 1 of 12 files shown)


build.gradle
```diff
@@ -27,12 +27,14 @@ repositories {
...
     compile 'org.slf4j:slf4j-api:1.7.25'
+    compileOnly 'ch.qos.logback:logback-classic:1.2.3'
     compileOnly 'javax.servlet:javax.servlet-api:3.1.0'
 
+    testCompile 'ch.qos.logback:logback-classic:1.2.3'
     testCompile 'junit:junit:4.12'
-    testCompile 'org.slf4j:slf4j-jdk14:1.7.25'
     testCompile 'javax.servlet:javax.servlet-api:3.1.0'
     testCompile 'org.mockito:mockito-core:2.10.0'
+    testRuntime 'org.codehaus.janino:janino:3.0.8'
```
fractalwrench: Is this dependency necessary?

lehphyro: It's used to make conditional logback configuration work, it's true it's not relevant for the implementation though.

lehphyro:
Do you have a date for when this is going to be in a release?

kattrali:
@lehphyro Not quite yet. There's a bit more work that we need to do on our end around the product parts of releasing this, like the documentation. There might be a few changes that we need to make as a part of that, though we can take it over the finish line from here as this is a great start. There should be an update soon though.

fractalwrench:
Apologies for leaving this for so long. We plan on releasing a Logback appender as part of #90, along with additional support for the Spring Framework.

PR #82, opened 2018-05-03
Next release

fractalwrench:


fractalwrench:
Generally most of the Java/mobile changelogs only reference PRs - maybe that's something we need to standardise across platforms?

PR #83, opened 2018-06-08
Session tracking on by default

fractalwrench:
## Goal

Enables automatic session tracking by default.

### Added

`setEndpoints` API on configuration class.

### Changed

- Auto capture enabled by default
- Deprecated `setEndpoint` and `setSessionEndpoint`

## Tests

Added and ran unit tests, as per the design spec.

## Review

  - [x] Final review

### DIFF
--- a/src/main/java/com/bugsnag/Configuration.java
+++ b/src/main/java/com/bugsnag/Configuration.java
@@ -10,6 +10,9 @@ import com.bugsnag.delivery.HttpDelivery;
 import com.bugsnag.delivery.SyncHttpDelivery;
 import com.bugsnag.serialization.Serializer;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -20,6 +23,8 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class Configuration {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(Bugsnag.class);
 
     private static final String HEADER_API_PAYLOAD_VERSION = "Bugsnag-Payload-Version";
     private static final String HEADER_API_KEY = "Bugsnag-Api-Key";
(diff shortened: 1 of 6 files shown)


src/main/java/com/bugsnag/Configuration.java
```diff
@@ -97,6 +102,52 @@ public boolean shouldAutoCaptureSessions() {
...
+            }
+        }
+
+        boolean invalidSessionsEndpoint = sessions == null || "".equals(sessions);
+        String sessionEndpoint = null;
+
+        if (invalidSessionsEndpoint) {
+            LOGGER.warn("The session tracking endpoint has not been"
+                    + " set. Session tracking is disabled");
+            this.autoCaptureSessions.set(false);
```
Pezzah: Should autoCaptureSessions also be disabled if setSessionEndpoint is called with empty / null?

src/main/java/com/bugsnag/Configuration.java
```diff
@@ -97,6 +102,52 @@ public boolean shouldAutoCaptureSessions() {
...
+        } else {
+            sessionEndpoint = sessions;
+        }
+
+        if (sessionDelivery instanceof HttpDelivery) {
+            ((HttpDelivery) sessionDelivery).setEndpoint(sessionEndpoint);
```
martin308: why does it set the endpoint here when it is possible that we have an invalid sessions endpoint? If we want to do this it might be worth adding a code comment as to why

src/main/java/com/bugsnag/Configuration.java
```diff
@@ -97,6 +102,52 @@ public boolean shouldAutoCaptureSessions() {
...
+            throw new IllegalArgumentException("Notify endpoint cannot be empty or null.");
+        } else {
+            if (delivery instanceof HttpDelivery) {
+                ((HttpDelivery) delivery).setEndpoint(notify);
+            } else {
+                LOGGER.warn("Delivery is not instance of HttpDelivery, cannot set notify endpoint");
+            }
+        }
+
+        boolean invalidSessionsEndpoint = sessions == null || "".equals(sessions);
```
martin308: is it better to use [`isEmpty()`](https://docs.oracle.com/javase/6/docs/api/java/lang/String.html#isEmpty())?

src/main/java/com/bugsnag/Configuration.java
fractalwrench: Resolved by in-person conversation - this is intended behaviour for the changeset

fractalwrench:
Addressed inline comments and added an extra test for the case when `Delivery` was implemented rather than `HttpDelivery`.

src/main/java/com/bugsnag/Configuration.java
```diff
@@ -97,6 +102,54 @@ public boolean shouldAutoCaptureSessions() {
...
+     * Note that if you are setting a custom {@link Delivery}, this method should be called after
+     * the custom implementation has been set.
+     *
+     * @param notify the notify endpoint
+     * @param sessions the sessions endpoint
+     *
+     * @throws IllegalArgumentException if the notify endpoint is empty or null
+     */
+    public void setEndpoints(String notify, String sessions) throws IllegalArgumentException {
+        if (notify == null || notify.length() == 0) {
```
martin308: is there a reason that we shouldn't be using [`isEmpty()`](https://docs.oracle.com/javase/6/docs/api/java/lang/String.html#isEmpty())?

src/main/java/com/bugsnag/Configuration.java
```diff
@@ -97,6 +102,54 @@ public boolean shouldAutoCaptureSessions() {
...
+            LOGGER.warn("The session tracking endpoint has not been"
+                    + " set. Session tracking is disabled");
+            this.autoCaptureSessions.set(false);
+        } else {
+            sessionEndpoint = sessions;
+        }
+
+        if (sessionDelivery instanceof HttpDelivery) {
+            // sessionEndpoint may be invalid (i.e. null) here, which should result in the default
+            // HttpDelivery throwing a MalformedUrlException which prevents delivery.
```
martin308: are we sure this is what we want to be doing in this case? It seems less optimal to use invalid data and rely on exceptions being thrown to prevent delivery

src/test/java/com/bugsnag/ConfigurationTest.java
```diff
@@ -40,4 +55,95 @@ public void testSessionApiHeaders() throws java.lang.Exception {
...
+            public void deliver(Serializer serializer, Object object, Map<String, String> headers) {
+            }
+
+            @Override
+            public void close() {
+            }
+        };
+        config.delivery = delivery;
+        config.sessionDelivery = delivery;
+        config.setEndpoints("http://example.com", "http://sessions.example.com");
```
martin308: do we need an assertion here?

fractalwrench:
@martin308 thanks for the comments. I've changed `SyncHttpDelivery` to fail faster if the endpoint is null, and read stderr to check that a message is logged when the session endpoint is invalid.

PR #85, opened 2018-07-25
refactor: fail build on compiler warnings

fractalwrench:
## Goal

Detect compiler warnings introduced in changes to Java code quickly.

## Changeset

- Updated gradle to error on any warnings found during compilation
- Address existing issues (raw types, missing `serialVersionUID`
- Suppress unchecked cast/deprecation warnings

## Tests

Ran tests on CI

## Review

  - [ ] Final review

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,3 +44,10 @@ test {
 checkstyle {
     toolVersion = "6.16"
 }
+
+gradle.projectsEvaluated {
+    tasks.withType(JavaCompile) {
+        // fail build on warnings, disable options complaining about Java 6 compatibility when building with JDK 7+
+        options.compilerArgs << "-Xlint:all" << "-Werror" << "-Xlint:-options"
+    }
+}
--- a/src/test/java/com/bugsnag/BugsnagTest.java
+++ b/src/test/java/com/bugsnag/BugsnagTest.java
@@ -154,11 +154,14 @@ public class BugsnagTest {
         Bugsnag bugsnag = new Bugsnag("apikey");
         bugsnag.setFilters("testfilter1", "testfilter2");
         bugsnag.setDelivery(new Delivery() {
+            @SuppressWarnings("unchecked")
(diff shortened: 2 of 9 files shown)


src/test/java/com/bugsnag/BugsnagTest.java
```diff
@@ -339,6 +342,7 @@ public void beforeNotify(Report report) {
         assertFalse(bugsnag.notify(new Throwable()));
     }
 
+    @SuppressWarnings("deprecation")
```
simonbowring: Probably worth a comment that it's the deprecated method that's under test. We should also add a test `testEndpoints` that tests `setEndpoints`.

build.gradle
```diff
@@ -44,3 +44,10 @@ test {
 checkstyle {
     toolVersion = "6.16"
 }
+
+gradle.projectsEvaluated {
+    tasks.withType(JavaCompile) {
+        // fail build on warnings, disable options complaining about Java 6 compatibility when building with JDK 7+
```
simonbowring: If we are supporting JDK 6 then I'm not sure that suppressing the Java 6 compatibility warning is something that we should do. If we want to guarantee compatibility then we should set `options.bootClasspath` to point to a JDK6 `rt.jar`.

On the other hand, I realize that this means that a JDK 6 installation is required to build which might not be practical. If you think it's fine to suppress this then we should look at getting JDK 6 CI sorted.

src/test/java/com/bugsnag/BugsnagTest.java
fractalwrench: `ConfigurationTest` covers `setEndpoints`, I'll update with an inline comment

build.gradle
fractalwrench: Fair point - I'll see if we can prioritise getting JDK 6 building on CI, or think about whether we need to support it still.

PR #86, opened 2018-07-27
Dockerise spring-web example project

fractalwrench:
Adds a Dockerfile that can be used to run the Spring Web example project:

```
docker build -t spring-web .
docker run -p 8080:8080 spring-web
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,9 +1,34 @@
 version: "3"
 services:
-  spring-web:
-    image: spring-web
+
+  servlet:
+    image: servlet
     ports:
-    - "8080:8080"
+      - "8080:8080"
     build:
       context: .
       dockerfile: Dockerfile
+    command: "./gradlew :examples:servlet:appRun"
+
+  simple:
+    image: simple
+    build:
+      context: .
+      dockerfile: Dockerfile
+    command: "./gradlew :examples:simple:run"
+
+  spring:
+    image: spring
+    build:
+      context: .
+      dockerfile: Dockerfile
+    command: "./gradlew :examples:spring:bootrun"
+
+  spring-web:
+    image: spring-web
+    ports:
+      - "8080:8080"
+    build:
+      context: .
+      dockerfile: Dockerfile
+    command: "./gradlew :examples:spring-web:bootrun"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,21 +7,18 @@ services:
       - "8080:8080"
     build:
       context: .
-      dockerfile: Dockerfile
     command: "./gradlew :examples:servlet:appRun"
 
   simple:
     image: simple
     build:
       context: .
-      dockerfile: Dockerfile
     command: "./gradlew :examples:simple:run"
 
   spring:
     image: spring
     build:
       context: .
-      dockerfile: Dockerfile
     command: "./gradlew :examples:spring:bootrun"
 
   spring-web:
@@ -30,5 +27,4 @@ services:
       - "8080:8080"
     build:
       context: .
-      dockerfile: Dockerfile
     command: "./gradlew :examples:spring-web:bootrun"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,9 @@
+version: "3"
+services:
+  spring-web:
+    image: spring-web
+    ports:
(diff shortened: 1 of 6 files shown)


fractalwrench:
Updated to add a docker-compose file that builds and runs the spring-web example project.

kattrali:
Do these files need to be in the root of the repository? What about when there are other examples to dockerise? The instructions for running with docker should also be added to the README in the example project.

fractalwrench:
I have updated the docker-compose file to add the remaining example apps, and have added a few quick notes on running with Docker to the README.

I believe leaving the Dockerfile in the root of the repository is required to build the example projects, as they have a dependency on the rootProject module. This also only requires one Dockerfile for all the example projects, as we can pass in the run command as a parameter when using `docker run` or `docker-compose` - is there a better alternative to this approach?

snmaynard:
This is reasonable. You can use build args in the dockerfile if there are a bunch of dependencies you want to install and dont want to have to install them each time you start the example (which might get old quick when developing).

We change the image built in the ruby maze tests which you can see at https://github.com/bugsnag/bugsnag-ruby/blob/master/features/fixtures/rails5/Dockerfile

fractalwrench:
@snmaynard I wasn't clear on whether this needed further changes to the Dockerfile or whether it's good to go - were there any further actions on this?

Changing the image via build args seems like a pretty cool feature. I think we're doing something similar for the nodejs mazerunner setup.

docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: "3"
+services:
+
+  servlet:
+    image: servlet
+    ports:
+      - "8080:8080"
+    build:
+      context: .
+    command: "./gradlew :examples:servlet:appRun"
```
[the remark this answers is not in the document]
martin308: Additionally it might be worth using the `gradle` base image. It was a lot faster in my testing.

```
image: gradle:jdk8
```

examples/servlet/README.md
```diff
@@ -16,3 +16,12 @@ A simple example application to show how Bugsnag works in a Servlet-based Java a
...
 
     http://localhost:8080/servlet
+
+## Running in Docker
+
+Execute the following commands in the root of the repository:
+
+```
+docker-compose build
+docker-compose up servlet
```
martin308: the container ended immediately for me after building this so I was unable to load the servlet. Do we need to do something to stop gradle exiting?

examples/spring-web/README.md
```diff
@@ -11,3 +13,12 @@ A Spring Boot example application to show how to use Bugsnag in a Spring web bas
...
 
     http://localhost:8080/
+
+## Running in Docker
+
+Execute the following commands in the root of the repository:
+
+```
+docker-compose build
+docker-compose up spring-web
```
martin308: this seems to work well except for the shutdown command which doesn't end the container. The output is below:

```
spring-web_1  | 2018-08-06 20:05:06.056  INFO 81 --- [nio-8080-exec-6] c.b.e.s.web.ApplicationRestController    : Shutting down application
spring-web_1  | 2018-08-06 20:05:06.057  INFO 81 --- [nio-8080-exec-6] ationConfigEmbeddedWebApplicationContext : Closing org.springframework.boot.context.embedded.AnnotationConfigEmbeddedWebApplicationContext@48ae9b55: startup date [Mon Aug 06 20:04:50 GMT 2018]; root of context hierarchy
spring-web_1  | 2018-08-06 20:05:06.058  INFO 81 --- [nio-8080-exec-6] o.s.j.e.a.AnnotationMBeanExporter        : Unregistering JMX-exposed beans on shutdown
spring-web_1  | 2018-08-06 20:05:06.059 DEBUG 81 --- [nio-8080-exec-6] com.bugsnag.Bugsnag                      : Closing connection to Bugsnag
```

examples/spring/README.md
```diff
@@ -7,3 +7,12 @@ A Spring Boot example application to show how to use Bugsnag in a Spring based J
...
     ../../gradlew clean bootRun
     ```
+
+## Running in Docker
+
+Execute the following commands in the root of the repository:
+
+```
+docker-compose build
+docker-compose up spring
```
martin308: this application never seems to end during my testing and has to be killed manually, is this expected?

PR #93, opened 2018-08-13
New severity reasons unhandledMiddlewareException and exceptionClass

simonbowring:
## Goal

In preparation for adding better support for Spring, this change adds two severity reasons useful for reporting exceptions from middleware.

## Design

Middleware, such as Spring, can automatically handle uncaught exceptions in order to render something meaningful to the end user, e.g. an HTTP `500` response to an API request that experienced an unexpected exception during processing. Such exceptions are not currently reported by the Bugsnag client but will be in the future. In preparation for this, adding an `unhandledExceptionMiddleware` severity reason to report these exceptions as unhandled with severity `error`. This severity reason is currently in use in other notifiers such as [bugsnag-ruby](https://github.com/bugsnag/bugsnag-ruby/blob/9f91115db8a37f67380205c0a7b6d0dda3fa2f83/lib/bugsnag/report.rb#L12).

Also adding a new severity reason `exceptionClass` that will allow particular exceptions to be assigned a specific severity. This is useful if a particular type of exception is deemed to be generally less important, e.g. an exception that results in a `400` response might only require a severity reason `info`.

### Added
Two new `SeverityReasonType`s:
  - `REASON_UNHANDLED_EXCEPTION_MIDDLEWARE("unhandledExceptionMiddleware")`
  - `REASON_EXCEPTION_CLASS("exceptionClass")`

A map `attributes` for `SeverityReason`.

## Tests

Added some tests to check that the new severity reasons and attributes are set correctly in the report payload.

## Review

  - [ ] Final review

### DIFF
--- a/src/main/java/com/bugsnag/HandledState.java
+++ b/src/main/java/com/bugsnag/HandledState.java
@@ -1,12 +1,16 @@
 package com.bugsnag;
 
+import java.util.Map;
+
 final class HandledState {
 
     enum SeverityReasonType {
         REASON_UNHANDLED_EXCEPTION("unhandledException"),
         REASON_HANDLED_EXCEPTION("handledException"),
         REASON_USER_SPECIFIED("userSpecifiedSeverity"),
-        REASON_CALLBACK_SPECIFIED("userCallbackSetSeverity");
+        REASON_CALLBACK_SPECIFIED("userCallbackSetSeverity"),
+        REASON_UNHANDLED_EXCEPTION_MIDDLEWARE("unhandledExceptionMiddleware"),
+        REASON_EXCEPTION_CLASS("exceptionClass");
 
         private final String name;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/bugsnag/HandledState.java
```diff
@@ -21,31 +25,53 @@ public String toString() {
     }
 
     private final SeverityReasonType severityReasonType;
+    private final Map<String, String> severityReasonAttributes;
```
simonbowring: I added the attributes in here to minimize the amount of refactoring, though it has made this class a little clunkier.

src/main/java/com/bugsnag/HandledState.java
```diff
@@ -73,6 +99,10 @@ SeverityReasonType calculateSeverityReasonType() {
                 : SeverityReasonType.REASON_CALLBACK_SPECIFIED;
     }
 
+    Map<String, String> getSeverityReasonAttributes() {
+        return severityReasonAttributes;
```
fractalwrench: We may want to consider wrapping this with `Collections.unmodifiableMap()` to make it very clear that this field shouldn't be mutated.

src/main/java/com/bugsnag/HandledState.java
```diff
@@ -21,31 +25,53 @@ public String toString() {
...
+    }
+
+    static HandledState newInstance(SeverityReasonType severityReason,
+                                    Map<String, String> severityReasonAttributes) {
+        return newInstance(severityReason, severityReasonAttributes, null, false);
     }
 
-    static HandledState newInstance(SeverityReasonType severityReasonType, Severity severity) {
+    static HandledState newInstance(SeverityReasonType severityReason, Severity severity) {
+        return newInstance(severityReason, null, severity, false);
```
fractalwrench: It'd be nice to consider passing in `Collections.emptyMap` so we can assume the field is always non-null. The getter could be annotated with a setter to prevent an empty object being serialised

simonbowring: We're already making sure that only non-empty fields are serialized:
https://github.com/bugsnag/bugsnag-java/blob/b17d7f3c7fee3e12a913f90bf86b7a947ea85ee4/src/main/java/com/bugsnag/serialization/Serializer.java#L20

PR #94, opened 2018-08-13
Initial restructure to add bugsnag-spring

simonbowring:
## Goal

In order to allow better support for spring, added a `bugsnag-spring` module. The end goal is to produce a bugsnag-spring JAR as well as the current one.

## Design

The `/src` directly was moved into its own module `bugsnag` in order to allow multiple modules. The Gradle build has been modified to build both modules. 

## Review

  - [ ] Final review


examples/spring/build.gradle
```diff
@@ -19,5 +19,5 @@ repositories {
 dependencies {
     compile("org.springframework.boot:spring-boot-starter")
 
-    compile rootProject
+    compile project(":bugsnag-java")
```
fractalwrench: Assuming at a later date we would switch to using bugsnag-spring here

bugsnag-spring/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+test {
+    testLogging {
+        events "passed", "skipped", "failed", "standardOut", "standardError"
+    }
+}
+
+checkstyle {
+    toolVersion = "6.16"
+    configFile = new File(rootDir, "config/checkstyle/checkstyle.xml")
+}
```
fractalwrench: It may be worth extracting some of this configuration to a [separate script](https://docs.gradle.org/current/userguide/plugins.html#sec:script_plugins), then applying it to both modules. That way it would only be necessary to update checkstyle/test/source compat in one place.

bugsnag-spring/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+
+sourceCompatibility = 1.6
+targetCompatibility = 1.6
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+}
```
fractalwrench: I think there are two things we might want to support here as a dependency:

1. the local version of the 'bugsnag-java' module
2. a mavenCentral artefact of bugsnag-java

I suppose the key question is which of those to use when we're making a release of bugsnag-spring. I think the advantages of using a maven artefact would be there's no potential for bugsnag-java to get out of date with bugsnag-spring fixes. Although that would mean we always have to release bugsnag-java first, and wait for it to propagate.

What were your thoughts on this?

build.gradle
```diff
@@ -16,31 +14,3 @@ buildscript {
 if (project.hasProperty('releasing')) {
     apply from: "release.gradle"
 }
```
fractalwrench: I think one approach here would be to apply `release.gradle` in both modules, and release each module as a separate artefact (with bugsnag-spring depending on bugsnag-java). It should be possible to update the script so that any module-specific values (e.g. package name) are supplied via `gradle.properties` or some equivalent mechanism.

In terms of testing this out I'd say we could mitigate the risk by:

- Triggering the task to generate a POM and manually inspecting it
- Uploading an alpha for both artefacts to JCenter/MavenCentral

Did you have this sort of approach in mind or something different?

examples/spring/build.gradle
simonbowring: Correct

bugsnag-spring/build.gradle
simonbowring: My personal preference would be to have the local version of the 'bugsnag-java' module as a dependency so that we don't have to wait. It would also make development of the bugsnag-spring module more straightforward when changes also need to be made to bugsnag-java.
You are right about bugsnag-java potentially getting out of date. I'm not sure of the best way around this.

build.gradle
simonbowring: This was the approach that I had in mind.

bugsnag-spring/build.gradle
fractalwrench: Another thing - if a project had a dependency on both bugsnag-java and bugsnag-spring, then this approach could potentially lead to some nasty issues with the classloader loading subtly different class definitions. I think this could be possible if someone misread our upgrade instructions, or if a Spring project depends on some 3rd party library that depends on bugsnag-java. I'm not sure that this scenario would be common enough to swing the argument one way or another.

We may be able to solve bugsnag-java getting out of date via tooling (e.g. checking the latest commit hash on master via a gradle plugin, before releasing bugsnag-spring).

simonbowring: How long does it usually take maven releases to propagate?

fractalwrench: 1-2 hours is typical, it can take longer however

simonbowring: What about something like this as a compromise:
```
dependencies {
    if (project.hasProperty('releasing')) {
        compile 'com.bugsnag:bugsnag:3.+'
    } else {
        compile project(':bugsnag-java')
    }
}
```

simonbowring: To keep consistent with the NDK rewrite approach, we are going to go with `compile project(':bugsnag-java')`.

.travis.yml
```diff
@@ -7,22 +7,22 @@ cache:
...
   - $HOME/.gradle/wrapper/
-install: ./gradlew :assemble
+install: ./gradlew :bugsnag:assemble :bugsnag-spring:assemble
 
 jobs:
   include:
   - stage: test
     jdk: openjdk8
-    script: ./gradlew check :test
+    script: ./gradlew check :bugsnag:test :bugsnag-spring:test
```
fractalwrench: this could be simplified to `./gradlew check test`

bugsnag-spring/build.gradle
```diff
@@ -0,0 +1,10 @@
+apply plugin: 'java-library'
+
+apply from: '../common.gradle'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
```
fractalwrench: it'd be good to add a dependency on the `bugsnag` module here

PR #95, opened 2018-08-14
Remove flaky session tracker test which intermittently fails on CI

fractalwrench:


fractalwrench:
I think it was just testing that creating a session 1000 times didn't exceed a timeout, which isn't a great test for CI.

PR #99, opened 2018-08-22
Spring Web MVC integration

simonbowring:
## Goal

Add better support for Spring Web MVC applications by:
- Automatically reporting uncaught errors thrown when processing HTTP requests as unhandled
- Automatically starting a Bugsnag session when an incoming request is received for Spring Boot (this already works for plain Spring applications)
- Automatically adding request metadata to handled and unhandled exception reports for Spring Boot (this already works for plain Spring applications)

### Added
- `BugsnagSpringConfiguration` class to automatically configure Bugsnag integration for Spring and Spring Boot applications.
- `BugsnagHandlerExceptionResolver` to report uncaught exceptions when processing HTTP requests
- `ErrorClassCallback` to set a custom severity if the reported exception is in a list of specific Spring exceptions 

### Changed
- `spring` and `spring-web` example apps now use `bugsnag-spring` 

## Tests

Tests are run against a Spring Boot 1.5 application that has been configured with Bugsnag by specifying `@Import(BugsnagSpringConfiguration.class)` only.

### Alternative Approaches

Spring Boot auto-configuration with a Spring Boot Starter was considered as a possible approach for automatic configuration but that would have only worked for Spring Boot applications. To keep a consistent approach for Spring Boot and plain Spring apps, a Spring configuration class was chosen that needs to be `@Import`ed into applications.

## Review

- [x] A changelog entry was added for the goal of this pull request
  - [ ] Final review


bugsnag/src/main/java/com/bugsnag/HandledState.java
```diff
@@ -11,7 +11,7 @@
         REASON_USER_SPECIFIED("userSpecifiedSeverity"),
         REASON_CALLBACK_SPECIFIED("userCallbackSetSeverity"),
         REASON_UNHANDLED_EXCEPTION_MIDDLEWARE("unhandledExceptionMiddleware"),
-        REASON_EXCEPTION_CLASS("exceptionClass");
+        REASON_ERROR_CLASS("errorClass");
```
simonbowring: Changing to reuse existing severity reason type used by other platforms.

bugsnag-spring/src/main/java/com/bugsnag/ErrorClassCallback.java
```diff
@@ -0,0 +1,77 @@
...
+        exceptionToSeverity.put(ServletRequestBindingException.class, Severity.INFO);
+        exceptionToSeverity.put(ConversionNotSupportedException.class, Severity.ERROR);
+        exceptionToSeverity.put(TypeMismatchException.class, Severity.INFO);
+        exceptionToSeverity.put(HttpMessageNotReadableException.class, Severity.WARNING);
+        exceptionToSeverity.put(HttpMessageNotWritableException.class, Severity.ERROR);
+        exceptionToSeverity.put(MethodArgumentNotValidException.class, Severity.INFO);
+        exceptionToSeverity.put(MissingServletRequestPartException.class, Severity.INFO);
+        exceptionToSeverity.put(BindException.class, Severity.INFO);
+        exceptionToSeverity.put(NoHandlerFoundException.class, Severity.INFO);
+        exceptionToSeverity.put(AsyncRequestTimeoutException.class, Severity.ERROR);
```
fractalwrench: it'd be good to test this out in a representative Spring application to determine whether we need to add any more. would be nice to get someone else from the pipeline team familiar with the ins and outs of Spring to check that these mappings seem reasonable, and we haven't missed any obvious ones out.

bugsnag-spring/src/main/java/com/bugsnag/SessionInterceptor.java
```diff
@@ -0,0 +1,38 @@
...
+    public boolean preHandle(HttpServletRequest request,
+                             HttpServletResponse response,
+                             Object handler) {
+
+        // If the request is available through the listener then the session has already started.
+        // This is only the case for plain Spring apps, ServletRequestListeners do not work for
+        // Spring Boot apps.
+        if (BugsnagServletRequestListener.getServletRequest() == null
+                && request.getDispatcherType() != DispatcherType.ERROR
+                && bugsnag.shouldAutoCaptureSessions()) {
```
fractalwrench: I'm not sure I follow 100% why we're accessing `getServletRequest` and `request` parameter here

bugsnag-spring/src/main/java/com/bugsnag/SpringWebMvcLoadedCondition.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+/**
+ * Check whether spring-webmvc is available to the application.
+ */
+class SpringWebMvcLoadedCondition implements Condition {
+    @Override
+    public boolean matches(ConditionContext context,
+                           AnnotatedTypeMetadata metadata) {
+        return context.getClassLoader().getResource("org/springframework/web/servlet") != null;
```
fractalwrench: `context.getClassLoader()` can return null

examples/spring-web/src/main/java/com/bugsnag/example/spring/web/Config.java
```diff
@@ -43,7 +46,6 @@ public String exampleWebsiteLinks() {
                 + "<a href=\"/send-handled-exception-info\">Send a handled exception to Bugsnag with INFO severity</a><br/>"
                 + "<a href=\"/send-handled-exception-with-metadata\">Send a handled exception to Bugsnag with custom MetaData</a><br/>"
                 + "<a href=\"/send-unhandled-exception\">Send an unhandled exception to Bugsnag</a><br/>"
-                + "<a href=\"/send-spring-handled-exception\">Send an exception handled by Spring @ExceptionHandler to Bugsnag</a><br/>"
```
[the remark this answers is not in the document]
fractalwrench: Whoops - very good point!

bugsnag-spring/src/main/java/com/bugsnag/RequestMetadataInterceptor.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Map;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Automatically adds metadata for HTTP requests which will be included in handled
+ * and unhandled reports generated during the processing of the request.
+ */
+class RequestMetadataInterceptor
```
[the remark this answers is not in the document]
simonbowring: Yes, I've added `springVersion` and optionally `springBootVersion`

bugsnag/src/test/java/com/bugsnag/HandledStatePayloadTest.java
```diff
@@ -131,8 +131,8 @@ public void testUnhandledMiddlewareSerialisation() throws java.lang.Exception {
     @Test
     public void testUnhandledExceptionClassSerialisation() throws java.lang.Exception {
         Report report = reportFromHandledState(HandledState.newInstance(
-                SeverityReasonType.REASON_EXCEPTION_CLASS,
-                Collections.singletonMap("exceptionClass", "TypeMismatchException"),
+                SeverityReasonType.REASON_ERROR_CLASS,
+                Collections.singletonMap("errorClass", "TypeMismatchException"),
```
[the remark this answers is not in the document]
simonbowring: I did the rename for convenience, but on second thought I think you're right, we should keep a distinction between error and exception. I'll change it back.

bugsnag-spring/src/main/java/com/bugsnag/RequestMetadataInterceptor.java
```diff
@@ -0,0 +1,63 @@
...
+
+            REQUEST_METADATA.set(map);
+        }
+        return true;
+    }
+
+    @Override
+    public void beforeNotify(Report report) {
+        Map<String, Object> map = REQUEST_METADATA.get();
+        if (map != null) {
```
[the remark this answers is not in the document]
simonbowring: Moved this section to `Report.java`

bugsnag/src/main/java/com/bugsnag/Bugsnag.java
```diff
@@ -550,6 +551,14 @@ public void close() {
         ExceptionHandler.disable(this);
     }
 
+    Configuration getConfig() {
+        return config;
+    }
+
+    SessionTracker getSessionTracker() {
```
simonbowring: Added to allow test access to session count

bugsnag/src/main/java/com/bugsnag/SessionTracker.java
```diff
@@ -69,6 +69,10 @@ Session getSession() {
         return session.get();
     }
 
+    SessionCount getBatchCount() {
```
simonbowring: Added to allow test access to session count

bugsnag/src/main/java/com/bugsnag/callbacks/ServletCallback.java
```diff
@@ -36,7 +36,8 @@ public void beforeNotify(Report report) {
         report
                 .addToTab("request", "url", request.getRequestURL().toString())
                 .addToTab("request", "method", request.getMethod())
-                .addToTab("request", "params", request.getParameterMap())
+                .addToTab("request", "params",
+                        new HashMap<String, String[]>(request.getParameterMap()))
```
simonbowring: This was required for testing because when the request context has ended, the parameter map is no longer available, even though the report object might still exist. Copying the context into a new list before this happens keeps the values available for the lifetime of the report object.

bugsnag/src/main/java/com/bugsnag/callbacks/ServletCallback.java
```diff
@@ -37,36 +34,12 @@ public void beforeNotify(Report report) {
                 .addToTab("request", "url", request.getRequestURL().toString())
```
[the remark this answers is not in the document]
simonbowring: No longer relevant as those classes no longer exist.

bugsnag-spring/src/main/java/com/bugsnag/BugsnagSpringConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+                public void beforeNotify(Report report) {
+                    report.addToTab("device", "springBootVersion", SpringBootVersion.getVersion());
+                }
+            };
+            bugsnag.addCallback(callback);
+            return callback;
+        }
+
+        /**
+         * Add a callback to add the version of Spring Boot used by the application.
```
kinbiko: Incorrect JavaDoc

bugsnag-spring/src/main/java/com/bugsnag/ExceptionClassCallback.java
```diff
@@ -0,0 +1,75 @@
...
+ * A callback to assign a specified severity in a report for particular Spring exceptions.
+ * This severity can be overridden if the severity is set manually in an application.
+ */
+class ExceptionClassCallback implements Callback {
+
+    private static final Map<Class<? extends java.lang.Exception>, Severity> exceptionToSeverity;
+
+    static {
+        exceptionToSeverity = new HashMap<Class<? extends java.lang.Exception>, Severity>();
+        exceptionToSeverity.put(HttpRequestMethodNotSupportedException.class, Severity.INFO);
```
kinbiko: Would be good to have comments for each of these explaining the reasoning behind the severity. At least for the non-error ones.

bugsnag-spring/src/main/java/com/bugsnag/SpringBootLoadedCondition.java
```diff
@@ -0,0 +1,22 @@
...
+ * Check whether spring-boot is available to the application.
+ */
+class SpringBootLoadedCondition implements Condition {
+
+    @Override
+    public boolean matches(ConditionContext context,
+                           AnnotatedTypeMetadata metadata) {
+        ResourceLoader loader = context.getResourceLoader() == null
+                ? new DefaultResourceLoader() : context.getResourceLoader();
+        return loader != null
```
kinbiko: The `loader` cannot be null here, and does not appear to be used. I think this method can be simplified to just

```java
return context.getClassLoader().getResource("org/springframework/boot") != null;
```
unless there's some magic going on on the line above. In case of magic, we should probably add a comment explaining what's going on.

This comment applies to other `Condition` implementations as well.

bugsnag-spring/src/test/java/com/bugsnag/SpringMvcTest.java
```diff
@@ -0,0 +1,291 @@
...
+        assertEquals("127.0.0.1", requestMetadata.get("clientIp"));
+
+        // Assert that the request params are as expected
+        @SuppressWarnings(value = "unchecked") Map<String, String[]> params =
+                (Map<String, String[]>) requestMetadata.get("params");
+        assertEquals("paramVal1", params.get("param1")[0]);
+        assertEquals("paramVal2", params.get("param2")[0]);
+
+        // Assert that the request headers are as expected
+        @SuppressWarnings(value = "unchecked") Map<String, String> headers =
```
kinbiko: Are we assuming that headers only have a single value? If I recall correctly, the HTTP standard allows multiple values to be set as values for headers.

bugsnag-spring/src/test/java/com/bugsnag/testapp/springboot/TestController.java
```diff
@@ -0,0 +1,79 @@
...
+     * Report a handled exception where the severity reason is exceptionClass
+     */
+    @RequestMapping("/handled-type-mismatch-exception")
+    public String handledTypeMismatchException() {
+        try {
+            throw new TypeMismatchException("Test", String.class);
+        } catch (TypeMismatchException ex) {
+            bugsnag.notify(ex);
+        }
+        return "The exception was handled";
```
kinbiko: Are these `return`s used anywhere? If not, might as well make the method `void`. Applies throughout.

bugsnag-spring/src/test/java/com/bugsnag/SpringMvcTest.java
simonbowring: I've added support for multiple headers by concatenating multiple values to a comma separated string.

bugsnag/src/main/java/com/bugsnag/callbacks/ServletCallback.java
```diff
@@ -61,10 +62,20 @@ private String getClientIp(HttpServletRequest request) {
...
+            StringBuilder value = new StringBuilder(headerValues.nextElement());
+
+            // If there are multiple values for the header, do comma-separated concat
+            // as per RFC 2616:
+            // https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
+            while (headerValues.hasMoreElements()) {
+                value.append(",").append(headerValues.nextElement());
+            }
+
+            map.put(key, value.toString());
```
fractalwrench: We should probably replace the value of certain headers with a placeholder (e.g. Authorisation, other sensitive keys).

bugsnag-spring/src/main/java/com/bugsnag/BugsnagHandlerExceptionResolver.java
```diff
@@ -0,0 +1,50 @@
+package com.bugsnag;
```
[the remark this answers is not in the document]
simonbowring: The reason for putting everything in the `com.bugsnag` package is so that we can add package-private middleware support to the `bugsnag` submodule for use in `bugsnag-spring` (and any middleware that we might want to support later) but don't want to add to the functionality to the public API, e.g. explicitly reporting unhandled exceptions.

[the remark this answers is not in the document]
fractalwrench: After in-person discussion I'm happy with this approach

bugsnag-spring/src/main/java/com/bugsnag/BugsnagSpringConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * Add a callback to add the version of Spring used by the application
+     */
+    @Bean
+    public Callback springVersionCallback() {
+        Callback callback = new Callback() {
+            @Override
+            public void beforeNotify(Report report) {
+                report.addToTab("device", "springVersion", SpringVersion.getVersion());
```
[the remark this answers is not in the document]
simonbowring: I don't think so, there aren't many repeated conditional checks at the moment and the conditionals are only run on startup during the building of the spring context, the performance impact seems negligible.

bugsnag/src/main/java/com/bugsnag/callbacks/ServletCallback.java
simonbowring: I've added Authorization to the filters as this is a registered header (see https://www.iana.org/assignments/message-headers/message-headers.xhtml). I'm not going to add any more at this stage as I think this needs wider discussion.

simonbowring:
@fractalwrench, I'll hold off updating the example readmes for now until we merge with the logback functionality which has a new way of initializing the Bugsnag client

PR #102, opened 2018-09-19
Spring async exception handling

simonbowring:
## Goal

Add better support for Spring async methods by:
- Adding a class to report uncaught exceptions thrown during the execution of async tasks.

**Note**
Async support needs to be added to an application manually, something like this (will be added to docs):
```
@Configuration
public class BugsnagAsyncConfig extends AsyncConfigurerSupport {
    @Autowired
    private Bugsnag bugsnag;

    @Override
    public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
        return new BugsnagAsyncExceptionHandler(bugsnag);
    }
}
```

Some investigation was made into whether it was possible to automatically add the config automatically, however multiple Spring `AsyncConfigurer`s cause the app to not start so this is not reliable.

### Added
`BugsnagAsyncExceptionHandler` The exception handler for uncaught exceptions during an async method call which notifies bugsnag of the exception

### Changed
- Rename `BugsnagHandlerExceptionResolver` to `BugsnagMvcExceptionHandler`
- Rename `BugsnagScheduledTaskErrorHandler ` to `BugsnagScheduledTaskExceptionHandler `

## Tests

Tests are run against a Spring Boot 1.5 application that has been configured with Bugsnag by specifying `@Import(BugsnagSpringConfiguration.class)` only.

## Review

  - [ ] Final review

### DIFF
--- a/bugsnag-spring/src/main/java/com/bugsnag/BugsnagAsyncExceptionHandler.java
+++ b/bugsnag-spring/src/main/java/com/bugsnag/BugsnagAsyncExceptionHandler.java
@@ -1,0 +1,38 @@
+package com.bugsnag;
+
+import com.bugsnag.HandledState.SeverityReasonType;
+
+import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
+import org.springframework.scheduling.annotation.AsyncConfigurer;
+import org.springframework.scheduling.annotation.AsyncConfigurerSupport;
+
+import java.lang.reflect.Method;
+import java.util.Collections;
+
+/**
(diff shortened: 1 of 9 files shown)


simonbowring:
I'm going to merge this down into the `spring-improvements` branch and raise a new PR when that branch is ready.

simonbowring:
@fractalwrench, as it turns out, unhandled exceptions in async methods that return a future **do** get reported, with the caveat that `@EnableScheduling` is used in the application. The scheduled task error handler handles it!

PR #104, opened 2018-09-27
Spring Notifier and Logback Appender

simonbowring:
## Goal

Add better support for Spring and Spring Boot applications by:
- Automatically reporting uncaught errors thrown when processing HTTP requests as unhandled
- Automatically starting a Bugsnag session when an incoming request is received for Spring Boot (this already works for plain Spring applications)
- Automatically adding request metadata to handled and unhandled exception reports for Spring Boot (this already works for plain Spring applications)
- Automatically reporting uncaught exceptions thrown during the execution of scheduled tasks
- Adding a class to report uncaught exceptions thrown during the execution of async tasks

**Note**
Async support needs to be added to an application manually:
```
@Configuration
public class BugsnagAsyncConfig extends AsyncConfigurerSupport {
    @Autowired
    private Bugsnag bugsnag;

    @Override
    public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
        return new BugsnagAsyncExceptionHandler(bugsnag);
    }
}
```
Some investigation was made into whether it was possible to automatically add the config automatically, however multiple Spring `AsyncConfigurer`s cause the app to not start so this is not reliable.

### Added
- `BugsnagSpringConfiguration` class to automatically configure Bugsnag integration for Spring and Spring Boot applications.
- `BugsnagMvcExceptionHandler` to report uncaught exceptions when processing HTTP requests
- `ExceptionClassCallback` to set a custom severity if the reported exception is in a list of specific Spring exceptions 
- `BugsnagScheduledTaskExceptionHandler` The error handler for uncaught exceptions on a scheduled task which:
  1. Notifies bugsnag of the exception
  1. Calls any error handler already configured by the application
- `BugsnagAsyncExceptionHandler` The exception handler for uncaught exceptions during an async method call which notifies bugsnag of the exception

Also adding a log appender to the java library so that errors can be sent to Bugsnag via Logback.

## Tests

Spring tests are run against a Spring Boot 1.5 application that has been configured with Bugsnag by specifying `@Import(BugsnagSpringConfiguration.class)`.

Tests also added to cover new log appender functionality and check the existing functionality works via the appender.

### Alternative Approaches

Spring Boot auto-configuration with a Spring Boot Starter was considered as a possible approach for automatic configuration but that would have only worked for Spring Boot applications. To keep a consistent approach for Spring Boot and plain Spring apps, a Spring configuration class was chosen that needs to be `@Import`ed into applications.

## Review

  - [ ] Final review


examples/spring-web/README.md
```diff
@@ -1,13 +1,15 @@
 # Bugsnag Spring web Java Example
 
-A Spring Boot example application to show how to use Bugsnag in a Spring web based Java application.
+Demonstrates how to use Bugsnag in a Spring Boot based Java application.
```
simonbowring: We should still mention that this is a Spring Boot web application

examples/spring/README.md
```diff
@@ -1,9 +1,15 @@
 # Bugsnag Spring Java Example
 
-A Spring Boot example application to show how to use Bugsnag in a Spring based Java application.
+Demonstrates how to use Bugsnag in a Spring based Java application.
```
simonbowring: We should clarify that this is a Spring Boot console application

CONTRIBUTING.md
```diff
@@ -22,6 +22,44 @@ Runs tests and checkstyle.
...
+   ```
+3. In your sample application `app/build.gradle`, add the following to the
+   dependencies section, inserting the exact version number required:
+
+   ```groovy
+   dependencies {
+     implementation 'com.bugsnag:bugsnag:[VERSION NUMBER]'
+   }
+   ```
+4. Clean your sample application and reload dependencies *every time* you
```
simonbowring: I didn't find that I needed to do this, at least when testing through intellij. Is this step definitely needed?

CONTRIBUTING.md
```diff
@@ -22,6 +22,44 @@ Runs tests and checkstyle.
...
+local repository and test the entire dependency flow inside of a sample
+application.
+
+To get started:
+
+1. In the `bugsnag-java` directory, run
+   `./gradlew -Preleasing=true publishToMavenLocal`.
+   This installs `bugsnag-java` and `bugsnag-spring` into your local
+   maven repository.
+2. In your sample application `build.gradle`, add `mavenLocal()` to the *top* of
```
simonbowring: Do we also need to supply instructions for Maven builds?

fractalwrench: As this is mainly intended for internal use, I think we can skip that for now

CONTRIBUTING.md
fractalwrench: I think it's worth doing just to be 100% sure that we're using the local artefact. I've previously experienced the gradle dependency cache on Android not picking up changes to the build script correctly, which this step should fix.

bugsnag/src/test/java/com/bugsnag/AppenderMetaDataTest.java
```diff
@@ -101,41 +100,30 @@ public void testMetaDataRemoval() {
...
-
-        // Send three test logs, the first one with report meta data added
-        LOGGER.warn("Test exception", new ExceptionWithCallback(new RuntimeException("test"),
-                new Callback() {
-                    @Override
-                    public void beforeNotify(Report report) {
-                        report.addToTab("report", "some key", "some report value");
-                    }
-                }));
+        BugsnagAppender.addThreadMetaData("report", "some key", "some report value");
```
snmaynard: this isnt required any more

CHANGELOG.md
```diff
@@ -1,5 +1,33 @@
...
+
+* Added `BugsnagAppender` that can report throwables from existing log statements to Bugsnag, using [logback](https://logback.qos.ch/)
+* [Spring] Automatically report exceptions thrown when processing MVC/REST requests
+* [Spring] Automatically report exceptions thrown in scheduled tasks
+* [Spring] Added `BugsnagAsyncConfig` class to simplify capture of uncaught exceptions in async tasks
+* [Spring] Automatically attach request metadata to reports
+* [Spring] Automatically attach Spring version information to reports
+* [Spring] Automatically track sessions for each MVC request
+
+**Deprecation notice**: `Bugsnag.init(...)` is now the preferred way to get an instance of the Bugsnag client
```
kattrali: Let's remove this deprecation and `Bugsnag.init()`, unless there was a technical reason to add it aside from the previous client sharing logic.

Pezzah: removed

PR #105, opened 2018-09-28
Update notifier payload info

fractalwrench:
## Goal

Updates the Notifier name in error/session payloads if the notifier is bugsnag-spring.

## Changeset

Adds a `NotifierUtils` class which checks whether the `BugsnagSpringConfiguration` class is available on the classloader. This is a good proxy for determining whether the notifier is `bugsnag-spring` or `bugsnag-java`.

## Tests

Added unit tests for serialisation, manual test of example plain/spring apps.

### DIFF
--- a/bugsnag-spring/build.gradle
+++ b/bugsnag-spring/build.gradle
@@ -13,6 +13,8 @@ repositories {
 
 dependencies {
     compile project(':bugsnag')
+    testCompile project(path: ':bugsnag', configuration: 'testRuntime')
+
     compileOnly "javax.servlet:javax.servlet-api:${servletApiVersion}"
     compileOnly "org.springframework:spring-webmvc:${springVersion}"
     compileOnly "org.springframework.boot:spring-boot:${springBootVersion}"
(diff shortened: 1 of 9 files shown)


bugsnag-spring/build.gradle
```diff
@@ -13,6 +13,8 @@ repositories {
 
 dependencies {
     compile project(':bugsnag')
+    testCompile project(path: ':bugsnag', configuration: 'testRuntime')
```
fractalwrench: This allows sharing of `BugsnagTestUtils` between the module's test code

bugsnag-spring/src/test/java/com/bugsnag/NotifierTest.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Test
+    public void testNotificationSerialisation() throws Throwable {
+        JsonNode payload = BugsnagTestUtils.mapReportToJson(config, this.report);
+        JsonNode notifier = payload.get("notifier");
+
+        assertEquals("Bugsnag Spring", notifier.get("name").asText());
+        assertEquals("https://github.com/bugsnag/bugsnag-java", notifier.get("url").asText());
+        assertNotNull(notifier.get("version").asText());
```
simonbowring: I would also check that the version is non-empty, same for sessions and the other NotifierTest class.

PR #110, opened 2018-10-10
fix: update autowiring to support spring 4.2.0

fractalwrench:
## Goal

Spring did not support constructor autowiring before 4.3, meaning we must autowire on fields instead (or inject dependencies manually through the constructor).

This changeset alters bugsnag-spring to autowire its configuration classes via fields, which prevents it crashing the application during initialisation.

## Changeset

- Separate configuration classes into separate source files
- Autowire via fields, not constructors

## Tests

Checked handled/unhandled, async, and scheduled task exceptions on Spring Boot 1.4.1 and Spring 4.2.

Note: Uncaught exceptions in async methods which return a Future do not appear to be captured by Bugsnag in Spring 4.2.

### DIFF
--- a/bugsnag-spring/src/main/java/com/bugsnag/SchedulingTaskConfiguration.java
+++ b/bugsnag-spring/src/main/java/com/bugsnag/SchedulingTaskConfiguration.java
@@ -23,14 +23,14 @@ class SchedulingTaskConfiguration implements SchedulingConfigurer {
     @Autowired
     private Bugsnag bugsnag;
 
-    private final BugsnagScheduledTaskExceptionHandler bugsnagErrorHandler =
-            new BugsnagScheduledTaskExceptionHandler(bugsnag);
-
     /**
      * Add bugsnag error handling to a task scheduler
      */
     @Override
     public void configureTasks(ScheduledTaskRegistrar taskRegistrar) {
+        BugsnagScheduledTaskExceptionHandler bugsnagErrorHandler =
+                new BugsnagScheduledTaskExceptionHandler(bugsnag);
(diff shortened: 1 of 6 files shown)


bugsnag-spring/src/main/java/com/bugsnag/SpringWebMvcConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.context.annotation.Configuration;
+
+import javax.annotation.PostConstruct;
+
+/**
+ * If spring-webmvc is loaded, add configuration for reporting unhandled exceptions.
+ */
+@Configuration
+@Conditional(SpringWebMvcLoadedCondition.class)
+class SpringWebMvcConfiguration {
```
simonbowring: Can we just call this class `MvcConfiguration`?

bugsnag-spring/src/main/java/com/bugsnag/SchedulingTaskConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.scheduling.config.ScheduledTaskRegistrar;
+import org.springframework.util.ErrorHandler;
+
+import java.lang.reflect.Field;
+
+/**
+ * Add configuration for reporting unhandled exceptions for scheduled tasks.
+ */
+@Configuration
+class SchedulingTaskConfiguration implements SchedulingConfigurer {
```
simonbowring: Can we call this class `ScheduledTaskConfiguration` instead?

PR #111, opened 2018-10-25
Move thread local metadata from BugsnagAppender to Bugsnag

simonbowring:
## Goal

Adding thread metadata on `Bugsnag` rather than `BugsnagAppender` allows the thread metadata to be added to `bugsnag.notify` reports and unhandled reports and log reports, not just the log reports.

In addition, allowing the thread metadata to be cleared with the following types of call manually:
- All metadata
- All metadata on a tab
- One item of metadata on a tab

These will need to be called manually in application code to clear down metadata as required on the thread, however all metadata on a thread will be cleared down automatically in the context of a servlet request when the servlet is destroyed.

## Design

Because metadata can be added globally, per thread and per report, there needs to be a precedence order for when there are duplicate keys in a tab. The precedence order (highest to lowest) is:

1. Report metadata
2. Thread metadata
3. Global metadata

## Tests

Mazerunner tests have been added to cover metadata added for:
- Log reports
- Handled reports
- Unhandled reports

The mazerunner tests cover metadata precedence and make sure that metadata from other threads does not bleed through.

There are currently only unit tests that cover thread metadata added with a log report, there should be more tests for `bugsnag.notify` and unhandled reports.

There should also be tests for servlet request thread metadata clearing, if possible. 

## Review

  - [ ] Final review


bugsnag/src/main/java/com/bugsnag/Bugsnag.java
```diff
@@ -601,6 +611,45 @@ public void close() {
...
+    // Thread metadata
+
+    /**
+     * Add a key value pair to a metadata tab just for this thread.
+     *
+     * @param tabName the name of the tab to add the key value pair to
+     * @param key     the key of the metadata to add
+     * @param value   the metadata value to add
+     */
+    public static void addThreadMetaData(String tabName, String key, Object value) {
```
fractalwrench: do we want to think about providing instance methods for this (maybe in addition to these existing methods)?

snmaynard: Why? These seem like static methods all day long to me.

fractalwrench: Most of the methods in our public Java API are instance methods. I felt it was worth raising the point to check that this feels consistent, and also to check that we're not making it static for all use-cases simply because we need to call this method in that way from within the notifier.

snmaynard: My preference is for methods that don't require an instance should be made static - but maybe that isnt a very java-ey thing to say. You can call a static method on an instance - but that is another can of worms https://www.quora.com/Why-cant-we-call-static-methods-using-the-object-in-Java

Pezzah:
Docs PR https://github.com/bugsnag/docs.bugsnag.com/pull/826

bugsnag/src/main/java/com/bugsnag/MetaData.java
```diff
@@ -15,6 +15,17 @@ public void clearTab(String tabName) {
         remove(tabName);
     }
 
+    public void clearKey(String tabName, String key) {
```
fractalwrench: this only needs to be package visible

bugsnag/src/test/java/com/bugsnag/ThreadMetaDataTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+                // Notify to Bugsnag
+                bugsnag.notify(new RuntimeException("test"));
+            }
+        });
+        thread.start();
+
+        // Wait for thread to run
+        try {
+            Thread.sleep(2000);
```
fractalwrench: The tests in this file currently take 6 seconds to run due to the use of `Thread.sleep`. We should rework these tests to use proper synchronisation mechanisms - for example, a `CountDownLatch`.

bugsnag/src/test/java/com/bugsnag/ThreadMetaDataTest.java
```diff
@@ -0,0 +1,185 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Map;
+
+public class ThreadMetaDataTest {
+
+    private static StubNotificationDelivery delivery;
+    private static Delivery originalDelivery;
+    private static Bugsnag bugsnag;
```
fractalwrench: Why do these fields need to be static?

bugsnag/src/test/java/com/bugsnag/ThreadMetaDataTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+import com.bugsnag.delivery.Delivery;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Map;
+
+public class ThreadMetaDataTest {
```
fractalwrench: It'd be good to add separate test methods for `clearThreadMetaData(String tabName)` and `clearThreadMetaData(String tabName, String key)`

bugsnag/src/main/java/com/bugsnag/BugsnagAppender.java
```diff
@@ -34,10 +31,6 @@
 /** Sends events to Bugsnag using its Java client library. */
 public class BugsnagAppender extends UnsynchronizedAppenderBase<ILoggingEvent> {
 
-    private static final String LOGGING_CONTEXT_THREAD_PREFIX
-            = "com.bugsnag.BugsnagAppender.thread.";
-    private static final String LOGGING_CONTEXT_TAB_SEPARATOR = ".reportTab.";
-
     /** Classes that we should not send logs for (to prevent infinite loops on error) */
     private static final List<String> EXCLUDED_CLASSES = Arrays.asList(
             "com.bugsnag.Bugsnag",
```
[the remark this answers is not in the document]
Pezzah: removed

bugsnag/src/main/java/com/bugsnag/MetaData.java
Pezzah: changed

bugsnag/src/test/java/com/bugsnag/ThreadMetaDataTest.java
Pezzah: removed static

bugsnag/src/test/java/com/bugsnag/ThreadMetaDataTest.java
Pezzah: changed to use `thread.join();`

PR #121, opened 2018-11-27
[PLAT-2110] Use deamon threads to start flushing sessions

Pezzah:
## Goal

In some cases the Bugsnag notifier can prevent clean application shutdowns due to the session flushing threads that are running.

This changes the threads to be marked as "deamon" so that the application can terminate without waiting for them to finish

The deamon threads run every 60 seconds to flush sessions created during that time. New threads are used to actually perform the flushing which are not marked as deamon so that they won't get killed part way though sending data.

## Review

  - [ ] Final review


bugsnag/src/main/java/com/bugsnag/util/DeamonThreadFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A version of {@link Executors#defaultThreadFactory()} which returns deamon threads
+ * This is to prevent applications from hanging waiting for the sessions scheduled task
+ * because deamon threads will be terminated on application shutdown
+ */
+public class DeamonThreadFactory implements ThreadFactory {
```
fractalwrench: The implementation of this class could be simplified by using the decorator pattern. We could create a field for `Executors.defaultThreadFactory()`, then make the thread returned by `threadFactory.newThread` a daemon before returning. That might also make a good candidate for a unit test.

bugsnag/src/main/java/com/bugsnag/util/DeamonThreadFactory.java
```diff
@@ -0,0 +1,54 @@
...
+
+        namePrefix = "deamon-pool-"
+                + poolNumber.getAndIncrement()
+                + "-thread-";
+    }
+
+    /**
+     * @see ThreadFactory#newThread(Runnable)
+     */
+    public Thread newThread(Runnable runner) {
```
fractalwrench: This is missing an override annotation

bugsnag/src/main/java/com/bugsnag/util/DeamonThreadFactory.java
```diff
@@ -0,0 +1,54 @@
...
+public class DeamonThreadFactory implements ThreadFactory {
+    private static final AtomicInteger poolNumber = new AtomicInteger(1);
+    private final ThreadGroup group;
+    private final AtomicInteger threadNumber = new AtomicInteger(1);
+    private final String namePrefix;
+
+    /**
+     * Constructor
+     */
+    public DeamonThreadFactory() {
```
fractalwrench: There's a typo in the class name (Deamon -> Daemon)

bugsnag/src/main/java/com/bugsnag/Bugsnag.java
```diff
@@ -81,7 +84,18 @@ private void scheduleSessionFlushes() {
         sessionExecutorService.scheduleAtFixedRate(new Runnable() {
             @Override
             public void run() {
-                sessionTracker.flushSessions(new Date());
+                // Create a new thread which is not a deamon thread to actually flush the sessions
+                // This will means that the thread will not be killed
+                // half way through if the application shuts down
+                Thread nonDeamonThread = new Thread(new Runnable() {
```
fractalwrench: We could use a cached executor e.g. `Executors.newSingleThreadExecutor()` rather than creating a new thread every time this is invoked.

bugsnag/src/main/java/com/bugsnag/util/DeamonThreadFactory.java
Pezzah: changed

bugsnag/src/main/java/com/bugsnag/util/DeamonThreadFactory.java
Pezzah: Added

bugsnag/src/main/java/com/bugsnag/util/DeamonThreadFactory.java
Pezzah: fixed

bugsnag/src/main/java/com/bugsnag/Bugsnag.java
Pezzah: Added new newSingleThreadExecutor

bugsnag/src/main/java/com/bugsnag/util/DeamonThreadFactory.java
Pezzah: also added a test

kattrali:
This issue came up before with a single thread executor in #50 ([current logic on master](https://github.com/bugsnag/bugsnag-java/blob/1ff0be9af090b0e4913160249c0952a1ff934d31/bugsnag/src/main/java/com/bugsnag/delivery/AsyncHttpDelivery.java#L21-L26)) - should the session flushing logic use an executor service in the same way to ensure the service doesn't hang an app? I'd just like to check that we aren't introducing a different way for the application to hang by fixing this one.

PR #122, opened 2018-11-28
[PLAT-2139] Ensure session counts are reported thread safe

Pezzah:
## Goal

Currently if there are multiple errors in quick succession, then the number of errors reported in the current session may be incorrect.

This will clone the current session when the metrics are updated to ensure that this cannot happen

## Review

  - [ ] Final review

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/Report.java
+++ b/bugsnag/src/main/java/com/bugsnag/Report.java
@@ -137,16 +137,16 @@ public class Report {
     void setSession(Session session) {
         if (session == null) {
             sessionMap = null;
+        } else {
+            sessionMap = new HashMap<String, Object>();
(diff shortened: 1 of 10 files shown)


bugsnag/src/main/java/com/bugsnag/Report.java
```diff
@@ -131,23 +131,22 @@ public String getContext() {
 
     @Expose
     Map<String, Object> getSession() {
+        return sessionMap;
+    }
+
+    void setSession(Session session) {
         if (session == null) {
-            return null;
+            sessionMap = null;
```
fractalwrench: We should add a unit test to verify that if the session parameter is null, the payload doesn't contain any session information. The current implementation will cause a crash due to NPEs.

PR #123, opened 2018-12-07
Allow thread meta data to be read

Pezzah:
## Goal

Currently it is not possible to read information from thread meta data out of the library, it is possible that someone would want to do this as part of the logging process.

The thread meta data should also be merged before the global callbacks are run, so that they can have access to a more complete report.

## Review

  - [ ] Final review

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/Bugsnag.java
+++ b/bugsnag/src/main/java/com/bugsnag/Bugsnag.java
@@ -446,6 +446,9 @@ public class Bugsnag {
             return false;
         }
 
+        // Add thread metadata to the report
+        report.mergeMetaData(THREAD_METADATA.get());
+
         // Run all client-wide beforeNotify callbacks
         for (Callback callback : config.callbacks) {
             try {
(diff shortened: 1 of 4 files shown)


bugsnag/src/main/java/com/bugsnag/Bugsnag.java
```diff
@@ -446,6 +446,9 @@ public boolean notify(Report report, Callback reportCallback) {
             return false;
         }
 
+        // Add thread metadata to the report
+        report.mergeMetaData(THREAD_METADATA.get());
```
simonbowring: This will have implications on which item of metadata is used in the case of duplicate keys when merging metadata. Before the precedence was:
1. Report metadata 
1. Thread metadata
1. Global metadata

Now it will be:
1. Report metadata
1. Global metadata
1. Thread metadata

Pezzah: That's correct. I think we should add the thread metadata first, so it can be used in any global level callbacks

fractalwrench:
Closing this PR as I can't see a ticket for it in the Pipeline Sprint and there hasn't been any further activity. @Pezzah or @simonbowring if there was a reason we need this in the notifier let me know and I can reopen.

Issue #120, opened 2018-11-20
Upgrade examples to a supported Spring Boot version


snicoll:
### Expected behavior
The samples use a supported version of Spring Boot.
### Observed behavior
The sample uses Spring Boot [1.4.1.RELEASE](https://github.com/bugsnag/bugsnag-java/blob/1ef85e5f6dd3045a6f3ae83af4dc7ca145c9bf52/examples/spring/build.gradle#L5) that is EOLed for quite a while. 1.5 is going to be EOLed [this August](https://spring.io/blog/2018/07/30/spring-boot-1-x-eol-aug-1st-2019).

PR #128, opened 2018-12-21
Use latest spring boot version for examples and tests

simonbowring:
## Goal
Use latest spring boot version in example apps as quite an old version was being used. Also, update maze fixtures to use same spring boot version.


fractalwrench:
Fixes #120

PR #131, opened 2019-01-16
Migrate non-standard device fields to metaData.device

fractalwrench:
## Goal

Migrates non-standard device fields which are not specified in our [Error API docs](https://bugsnagerrorreportingapi.docs.apiary.io/#reference/0/notify/send-error-reports) to `metaData.device` instead.

## Changeset

- Migrated `osArch`, `runtimeName`, `runtimeVersion`, and `locale` to `metaData.device`.
- Deprecated `report.setAppInfo` and `report.setDeviceInfo` in favour of `report.addToTab`

## Tests

Ran existing tests on CI, inspected payload manually to confirm that the JSON appears as expected, and confirmed that each field still shows on handled/unhandled errors in the dashboard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.X.X (TBD)
+
+* Migrate non-standard device fields to metaData.device
+  [#131](https://github.com/bugsnag/bugsnag-java/pull/131)
+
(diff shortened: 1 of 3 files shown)


fractalwrench:
@duncanhewett I'm assuming you're happy with the deprecation of these two methods in favour of using `addToTab`?

PR #137, opened 2019-03-11
Merge internal checkstyle rules

simonbowring:
In preparation for hosting a common checkstyle config for all projects, this PR uses the existing bugsnag-java checkstyle config as a base and merges in additional rules used for internal projects.

The only existing rule that has been explicitly modified is upping the max line length from 100 to 120.


bugsnag-spring/src/test/java/com/bugsnag/SpringMvcTest.java
```diff
@@ -202,6 +202,7 @@ public void beforeNotify(Report report) {
         } finally {
             // Remove the callback via reflection so that subsequent tests do not use it
             Field callbacksField = Configuration.class.getDeclaredField("callbacks");
+            callbacksField.setAccessible(true);
```
fractalwrench: Why is this change necessary?

bugsnag/src/main/java/com/bugsnag/Configuration.java
```diff
@@ -30,21 +30,21 @@
     private static final String HEADER_API_KEY = "Bugsnag-Api-Key";
     private static final String HEADER_BUGSNAG_SENT_AT = "Bugsnag-Sent-At";
 
-    public String apiKey;
-    public String appType;
-    public String appVersion;
-    public Delivery delivery = new AsyncHttpDelivery(SyncHttpDelivery.DEFAULT_NOTIFY_ENDPOINT);
-    public Delivery sessionDelivery =
+    private String apiKey;
```
fractalwrench: This is a breaking change - these fields need to remain public.

bugsnag-spring/src/test/java/com/bugsnag/SpringMvcTest.java
simonbowring: Because callbacks were changed from public to private. This can be reverted if we're going to back out the accessors.

Issue #132, opened 2019-01-31
Bugsnag-Spring & logback


sraveh:
Hi,

Im trying to use bugsnag-spring 3.4.4 with spring boot 2.1.0. Im using log4j2 and not logback. Logback is excluded and not on the classpath. It seems that bugsnag-java is dependent on logback. Im getting the following error when the application start:

caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'com.bugsnag.BugsnagSpringConfiguration': Invocation of init method failed; nested exception is java.lang.NoClassDefFoundError: ch/qos/logback/core/UnsynchronizedAppenderBase

fractalwrench:
Thanks for the report @sraveh - this has been fixed in v3.4.6.

PR #140, opened 2019-04-12
Swallow Throwables thrown when configuring bugsnag appender

fractalwrench:
## Goal
Not all users use Logback in their project, so `BugsnagSpringConfiguration` calling `BugsnagAppender` methods can throw a `NoClassDefFoundError`, as observed in #132. If the dependency is not on the classpath, we should ignore this error and continue initialising bugsnag as normal.

## Tests
Ran the example app with [logback excluded](https://github.com/bugsnag/bugsnag-java/compare/log4j-example?expand=1), and confirmed that this change allowed the application to startup.

### DIFF
--- a/bugsnag-spring/src/main/java/com/bugsnag/BugsnagSpringConfiguration.java
+++ b/bugsnag-spring/src/main/java/com/bugsnag/BugsnagSpringConfiguration.java
@@ -2,7 +2,6 @@ package com.bugsnag;
 
 import com.bugsnag.callbacks.Callback;
 
-import org.springframework.beans.factory.UnsatisfiedDependencyException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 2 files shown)


bugsnag-spring/src/main/java/com/bugsnag/BugsnagSpringConfiguration.java
```diff
@@ -48,21 +48,25 @@ ScheduledTaskBeanLocator scheduledTaskBeanLocator() {
...
-        BugsnagAppender.addExcludedLoggerPattern("org.eclipse.jetty.server.HttpChannel");
+            // Exclude Jetty logger when processing HTTP requests via the HttpChannel
+            BugsnagAppender.addExcludedLoggerPattern("org.eclipse.jetty.server.HttpChannel");
 
-        // Exclude Undertow logger when processing HTTP requests
-        BugsnagAppender.addExcludedLoggerPattern("io.undertow.request");
+            // Exclude Undertow logger when processing HTTP requests
+            BugsnagAppender.addExcludedLoggerPattern("io.undertow.request");
+        } catch (Throwable ignored) {
+        }
```
kattrali: 🚗 ))) Drive-by review: There should be a comment indicating why an exception is ignored here. Also, can the `Throwable` be a more specific subclass?

kinbiko: I may be wrong, but after we do this I think we'd be able to remove the `@SuppressWarnings` annotation. I think a comment is sufficiently non-empty.

PR #141, opened 2019-04-17
Migrate version information to device.runtimeVersions

fractalwrench:
## Goal

We should collect the following information in the `device.runtimeVersions` section of the payload:

- Java version
- Java type (runtime environment)
- Spring version
- Spring Boot version

## Changeset

- Moved `device.runtimeName` to `device.runtimeVersions.javaType`
- Moved `device.runtimeVersion` to `device.runtimeVersions.javaVersion`
- Moved `device.metaData.springVersion` to `device.runtimeVersions.springFramework`
- Moved `device.metaData.springBootVersion` to `device.runtimeVersions.springBoot`
- Added package-visible `BeforeSendSessions` callback which is invoked immediately before session delivery and allows mutation of the payload
- Added Spring/Spring Boot session callbacks that add runtime version info for sessions

## Tests

Ran existing tests, and added new mazerunner scenarios to verify that version info is included in both errors/sessions for a plain, Spring, and Spring Boot app.

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/Diagnostics.java
+++ b/bugsnag/src/main/java/com/bugsnag/Diagnostics.java
@@ -45,4 +45,17 @@ class Diagnostics {
         }
         return map;
     }
+
+    @SuppressWarnings("unchecked")
+    static Map<String, Object> retrieveRuntimeVersionsMap(Map<String, Object> device) {
+        Object obj = device.get("runtimeVersions");
+
+        if (obj instanceof Map) {
+            return (Map<String, Object>) obj;
+        } else { // fallback to creating a new map if payload was mutated
+            Map<String, Object> map = new HashMap<String, Object>();
+            device.put("runtimeVersions", map);
+            return map;
+        }
+    }
 }
--- a/bugsnag/src/main/java/com/bugsnag/Diagnostics.java
+++ b/bugsnag/src/main/java/com/bugsnag/Diagnostics.java
@@ -23,7 +23,15 @@ class Diagnostics {
         map.put("hostname", DeviceCallback.getHostnameValue());
         map.put("osName", System.getProperty("os.name"));
         map.put("osVersion", System.getProperty("os.version"));
+        map.put("runtimeVersions", getRuntimeVersions());
         return map;
+    }
+
+    private Map<String, String> getRuntimeVersions() {
+        Map<String, String> runtimeVersions = new HashMap<String, String>();
+        runtimeVersions.put("javaType", System.getProperty("java.runtime.name"));
+        runtimeVersions.put("javaVersion", System.getProperty("java.runtime.version"));
+        return runtimeVersions;
     }
 
     private Map<String, Object> getDefaultAppInfo(Configuration configuration) {
--- a/bugsnag/src/main/java/com/bugsnag/Diagnostics.java
+++ b/bugsnag/src/main/java/com/bugsnag/Diagnostics.java
@@ -47,15 +47,16 @@ class Diagnostics {
     }
 
     @SuppressWarnings("unchecked")
-    static Map<String, Object> retrieveRuntimeVersionsMap(Map<String, Object> device) {
+    static void addDeviceRuntimeVersion(Map<String, Object> device, String key, Object value) {
         Object obj = device.get("runtimeVersions");
+        Map<String, Object> runtimeVersions;
 
         if (obj instanceof Map) {
-            return (Map<String, Object>) obj;
+            runtimeVersions = (Map<String, Object>) obj;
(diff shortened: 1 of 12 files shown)


bugsnag/src/main/java/com/bugsnag/Diagnostics.java
```diff
@@ -37,4 +45,17 @@
         }
         return map;
     }
+
+    @SuppressWarnings("unchecked")
+    static Map<String, Object> retrieveRuntimeVersionsMap(Map<String, Object> device) {
+        Object obj = device.get("runtimeVersions");
+
+        if (obj instanceof Map) {
+            return (Map<String, Object>) obj;
```
bengourley: If we're being paranoid that this was replaced with something that isn't a `Map`, shouldn't we also be paranoid about it being a `Map<K, V>` where `K` is not `String` or `V` is not `Object`?

And seeing as this static method's job is to mutate the `device` property of a report, wouldn't this functionality be better suited to a mutator method on a report class itself?

```java
public Report addDeviceRuntimeVersion(String key, Object value) {
    Object obj =  diagnostics.device.get("runtimeVersions");
    // do the paranoid thing here until we have a Map<String, Object> for runtimeVersions
    Map<String, Object> runtimeVersions;
    if (/* can you check if it is a Map<String, Object> ? */) {
        runtimeVersions  = (Map<String, Object>) obj;
    } else {
        runtimeVersions = new HashMap<String, Object>();
    }
    runtimeVersions.put(key, value);
    return this;
}
```

fractalwrench: This method mutates a `device` map for both the `Report` and `SessionPayload` class, therefore the `Diagnostics` class felt like the best place.

Unfortunately it's not possible in Java to check at runtime whether an `Object` is of type `Map<String, Object>`, and the best we can do is check that is of type `Map`.

The `<String, Object>` syntax specifies [generic type parameters](https://docs.oracle.com/javase/tutorial/java/generics/index.html), which says that the key for the map will be of type `String`, and the value for the map will be of type `Object`. Because these are specified generically, we could reuse the same Map class with different types (e.g. a type of `Foo` for a key, and `Bar` for a value. These generic types are enforced at compile-time and will lead to an error if a wrong type is supplied, for example:

```
Map<String, Object> map = new HashMap<>();
map.put("test", someObj); // compiles, because both params are correct types
map.put(someObj, "test"); // fails at compile-time as 1st param is not of type String
```

Unfortunately at runtime, this type information [is erased](https://stackoverflow.com/questions/339699/java-generics-type-erasure-when-and-what-happens) (partly for backward compatibility). This means that when our example above is compiled, it would look more like this:

```
Map map = new HashMap();
map.put("test", someObj);
map.put(someObj, "test");
```

Therefore the best we can do is use `instanceof Map`, which verifies an object is of type `Map`, then cast the class to `Map<String, Object>` to regain the generic information (this is why `@SuppressWarnings("unchecked")`) is required.

bengourley: Ok – understood re generic type erasure, I wondered if this might be the case. Thanks for explaining!

Also undertstood re this method being used for both sessions and reports, so I agree it should go in the Diagnostics class.

The current changeset still seems a little clunky to me, and I think it could be improved in terms of more terseness, less indirection and more clarity.

```java
class Diagnostics {
  static void addDeviceRuntimeVersion(Map<String, Object> device, String key, Object value) {
    // get runtime versions map implementation as above
    runtimeVersions.put(key, value)
  }
}
```
Then in the spring configuration class

```diff
    @Bean
    BeforeSendSession springVersionSessionCallback() {
        BeforeSendSession beforeSendSession = new BeforeSendSession() {
            @Override
            public void beforeSendSession(SessionPayload payload) {
-              appendSpringRuntimeVersion(Diagnostics.retrieveRuntimeVersionsMap(payload.getDevice()));
+              addSpringRuntimeVersion(payload.getDevice());
            }
        };
        bugsnag.addBeforeSendSession(beforeSendSession);
        return beforeSendSession;
    }

-   private void appendSpringRuntimeVersion(Map<String, Object> runtimeVersions) {
+   private void addSpringRuntimeVersion(Map<String, Object> device) {
-       runtimeVersions.put("springBoot", SpringBootVersion.getVersion());
+       Diagnostics.addDeviceRuntimeVersion(device, "springFramework", SpringVersion.getVersion());
    }
```

To me this communicates "I'm adding the spring version to the device runtime versions map" more clearly, but we're getting into subjective/style territory here, so this is _food for thought_ rather than _changes required_.

fractalwrench: That approach looks much nicer, I'll update the PR.

PR #147, opened 2019-07-03
Allow a BugsnagAppender to be created from an existing client.

safarmer:
This makes it possible to programatically create a Bugsnag logback appender without having to pass in the configuration options twice (once to the original client, and once again to the appender or defined in the logback XML file). This is especially useful if you are adding callbacks and configuration settings that are the same across the original instance and the Logback appender.

If the appender is created with an existing client, it is not possible to set any of the values of the appender that affect Bugsnag as it would also affect the original client.

## Goal
Make it easier to share the configuration and callbacks between an existing Bugsnag client and a Bugsnag Logback appender.

## Design
This approach was taken to couple the logback and client configurations when both are used. This can be the case when you are using Spring Boot.

### Changed
The BugsnagAppender has an optional constructor with a Bugsnag client argument that can be used as the client for logging.

## Tests
Added a new JUnit test to ensure that the new client is used and not overwritten when the appender is started.

### Alternative Approaches

N/A

### Outstanding Questions

Should we explicitly throw an exception if calling a method on the appender that would change the Bugsnag client configuration when providing a client?

Should the provided client be able to be reconfigured through the appender? This sounds dangerous to me, and the client will only ever be provided if creating the appender programatically.

### Linked issues
N/A

## Review

  - [x] Final review

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/BugsnagAppender.java
+++ b/bugsnag/src/main/java/com/bugsnag/BugsnagAppender.java
@@ -80,9 +80,26 @@ public class BugsnagAppender extends UnsynchronizedAppenderBase<ILoggingEvent> {
     /** Bugsnag client. */
     private Bugsnag bugsnag = null;
 
+    /** Creates an appender from an existing Bugsnag client. */
+    public BugsnagAppender(Bugsnag bugsnag) {
+        super();
+        this.bugsnag = bugsnag;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


tomlongridge:
Thanks @safarmer, we'll take a look at this and get back to you shortly.

safarmer:
Thanks for the quick review @fractalwrench! Looking forward to seeing this in an upcoming release.

fractalwrench:
This has been shipped in v3.6.0. Thanks again for your contribution!

PR #153, opened 2020-01-27
Make FilteredMap key matching case-insensitive

### LINKED ISSUE #152: Request tab includes cookies
### Expected behavior
Cookies are filtered from request headers metadata

### Observed behavior
`Request` tab contains a `headers` map with `cookie` present.

### Steps to reproduce
We’re using `bugsnag-spring` with Spring Boot 2.0.9 and Tomcat 9.0.30.

### Version
3.6.1

### Additional information
It looks like the issue is that the underlying servlet implementation stores header keys case-insensitively, so the default `Cookie` filter does not match. When it comes to filtering, I think it would make sense to make all comparisons case-insensitive, so that a filter for `"password"` matches `"USER_PASSWORD"`, `"adminPassword"`, etc.

eager:
# Goal

`FilteredMap` is currently case-sensitive, making complete matching tedious. For example, to match the keys `password`, `USER_PASSWORD`, and `adminPassword` would require three separate filters, `password`, `PASSWORD`, and `Password`. Filtering becomes simpler, and less surprising, if matching is case-insensitive.

### Changed

`FilteredMap#shouldFilterKey` converts both the key and filter to lower case before matching.

## Tests

`FilteredMapTest` has additional test cases.

### Alternative Approaches

None

### Outstanding Questions

None

### Linked issues

Fixes #152

## Review

  - [ ] Final review

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/util/FilteredMap.java
+++ b/bugsnag/src/main/java/com/bugsnag/util/FilteredMap.java
@@ -113,7 +113,7 @@ public class FilteredMap implements Map<String, Object> {
         }
 
         for (String filter : keyFilters) {
-            if (key.contains(filter)) {
+            if (key.toLowerCase().contains(filter.toLowerCase())) {
                 return true;
             }
         }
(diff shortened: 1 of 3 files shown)


phillipsam:
Hi @eager - thanks for the PR. We would be interested in making this configuration option capable of taking regexes. If you were interested in making a PR for this, we'd be happy to take a look.

PR #157, opened 2020-11-06
fix: jvm hangs when System.exit or bugsnag.close is not called

### LINKED ISSUE #151: Bugsnag prevents JVM from shutting down normally
### Expected behavior
When all non daemon threads of any app  finish, the JVM shuts down normally... After implementing Bugsnag JVM still runs. This behavior is not what I expect, especially for AWT/Swing apps.

### Steps to reproduce
`public class Test {

    public static final void main(String[] args) {
        Bugsnag instance = new Bugsnag("api key");
    }
}`
JVM will never shut down...

### Version
3.6.1

### Additional information
If this cannot be fixed, please update your implementation guide to call System.exit(0) or bugsnag.close() somewhere in the code. Calling System.exit(0) is not common in AWT/Swing apps as user closes window and UI thread finishes normally.

This issue made our app update mechanism broken as JVM was running and old  jar couldn't be overwritten with new one. Also it made our app unusable after broken update as we use FileLock to enable single instance of app and and lock was never released as JVM kept running. This has broken around 100 installations of our app before we realized what has happened and stopped update. Unfortunately we didn’t discover this bug during our testing, only after it was released to production.

tomlongridge:
## Goal

The fix to prevent a resource leak in #143 improved the shutdown hook to prevent the resource leak (calling `ExceptionHandler.disable` and `config.delivery.close`) but also reverted the task scheduler from running as a daemon thread. Therefore, in a console app, or similar, which doesn't explicitly call `System.exit`, the JVM waits indefinitely for the `sessionExecutorService` to stop before triggering the shutdown hook.

This PR should be the benefits of both #143 and #121 to gracefully shutdown in both use cases.

Fixes #151 

## Changeset

* Reinstated the `DaemonThreadFactory` (removed in #143)
* Removed `System.exit` calls from the Spring Boot example as the explicit call should not be required.

## Testing

* Updated the "mazerunner" (Spring boot app) end-to-end test to remove the `System.exit` call to test this in future.
* Also ran the fix in a standalone Tomcat war to ensure the warnings about resource leaks were not seen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## TBD
+
+* Fix JVM hang when System.exit or bugsnag.close is not called
+  [#157](https://github.com/bugsnag/bugsnag-java/pull/157)
+
 ## 3.6.1 (2019-15-08)
 
 * Prevent potential ConcurrentModificationException when adding callback
--- a/bugsnag/src/main/java/com/bugsnag/Bugsnag.java
+++ b/bugsnag/src/main/java/com/bugsnag/Bugsnag.java
@@ -3,6 +3,7 @@ package com.bugsnag;
 import com.bugsnag.callbacks.Callback;
 import com.bugsnag.delivery.Delivery;
 import com.bugsnag.delivery.HttpDelivery;
+import com.bugsnag.util.DaemonThreadFactory;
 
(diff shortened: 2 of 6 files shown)


fractalwrench (approved):

nullobject:
Nice :clap:

PR #163, opened 2021-04-29
Add license audit to CI

Cawllec:
## Goal

Uses buildkite to verify all dependency licenses are appropriate for use


Cawllec:
To reconcile the differences between branches I'm going to close this, and reopen it when fixed

PR #166, opened 2021-07-20
Suggested change to fix logging of handled exceptions

### LINKED ISSUE #156: Fix redundant exception handling
### Describe the bug

When an unhandled exception occurs, the Spring framework iterates through a list of exception handlers looking for someone to consume the unhandled exception. [This](https://github.com/spring-projects/spring-framework/blob/6fbd4841ec4a20cfd8c5338c6bcf0171839c2326/spring-webmvc/src/main/java/org/springframework/web/servlet/DispatcherServlet.java#L1299) is the Java for loop in the Spring framework that is looking for 1 and only one exception handler to swallow the exception.

The existing BugSnag code explicitly sets `BugsnagMvcExceptionHandler` as the first Exception Handler by setting `Ordered.HIGHEST_PRECEDENCE` on line 22. The code on line 47-48 then explicitly lets some other exception handler try to consume the exceptions. This breaks the expectation that only 1 handler should be invoked.

It's more reasonable that the BugSnag library should run last in the chain and not first in the chain. Allowing other exception handlers to consume the exception first allows the application to consume or 'handle' the exception and swallow it before it reaches BugSnag.

For example, my Java application throws custom defined exceptions to perform URL redirects and display custom HTML error pages. I have a @ControllerAdvice class that then properly renders the content. However, the BugSnag library catches the exception, logs it to your API, and then finally my application code runs. Since I handled the exception, it should never have been logged by the library.

Another example, in my class with custom exception handling using @ControllerAdvice, for 1 particular exception, I explicitly invoke BugSnag to log an exception as handled. But then your code also logs it as unhandled in your own handler..

### Steps to reproduce

Create a spring boot project
Create a class to handle a custom exception and create a @ControllerAdvice annotated class to custom handle the exception.
Invoke the exception and watch BugSnag log it as unhandled as well as watch the @ControllerAdvice annotated class handle the exception.

### Environment
* Bugsnag version:3.6.1
* Java version: 8
* Spring framework version (if any): 2.1.4.RELEASE
* Maven version (if any): 3.6.3

### Example

I created this branch to solve the issue, but I couldn't understand the test that I broke. This example fix and the above referenced parts of the code explain well what the observed yet expected result should be.

https://github.com/bugsnag/bugsnag-java/compare/master...nkavian:lower-precedence?expand=1

Change the order on BugsnagMvcExceptionHandler from `highest` to `lowest`. This allows the application to handle exceptions. If an exception truly was not handled by any one else, then finally it should be logged to BugSnag as unhandled.

Using `bugsnag.setIgnoreClasses()` is a workaround but it's undesirable as a long term solution. I have to maintain a list of 10+ classes to exclude here.

TobiasMende:
I had an idea how to fix #156 and would love some feedback on this approach. If the approach looks okay to you, I would spend some more time to add tests to verify the changed behaviour (namely `@ResponseStatus` exceptions and `@ExceptionHandler` methods get to handle the exception first and therefore can prevent BugSnag from logging it).

## Goal

Fix #156 

## Design

Just changing the Order of the `BugsnagMvcExceptionHandler` does not work, because all the three exception handlers have the same (lowest) precedence but are grouped in the `HandlerExceptionResolverComposite`.  This approach injects BugSnag as the second last exception handler just before the `DefaultHandlerExceptionResolver`.

## Changeset

The `BugsnagMvcExceptionHandler` is injected manually at the right position and not by Spring ordering.

## Testing

Existing tests still pass. I added two new test to test that exceptions handled by `@ResponseStatus` and `@ExceptionHandler` are not sent to BugSnag.

### DIFF
--- a/bugsnag-spring/src/main/java/com/bugsnag/MvcConfiguration.java
+++ b/bugsnag-spring/src/main/java/com/bugsnag/MvcConfiguration.java
@@ -4,7 +4,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.web.servlet.HandlerExceptionResolver;
-import org.springframework.web.servlet.config.annotation.WebMvcConfigurationSupport;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 
 import java.util.List;
 import javax.annotation.PostConstruct;
@@ -14,7 +14,7 @@ import javax.annotation.PostConstruct;
  */
 @Configuration
 @Conditional(SpringWebMvcLoadedCondition.class)
-class MvcConfiguration extends WebMvcConfigurationSupport {
+class MvcConfiguration extends WebMvcConfigurerAdapter {
 
     @Autowired
     private Bugsnag bugsnag;
@@ -41,7 +41,7 @@ class MvcConfiguration extends WebMvcConfigurationSupport {
      * Only unhandled exceptions shall be sent to Bugsnag.
      */
     @Override
-    protected void extendHandlerExceptionResolvers(List<HandlerExceptionResolver> exceptionResolvers) {
+    public void extendHandlerExceptionResolvers(List<HandlerExceptionResolver> exceptionResolvers) {
         final int position = exceptionResolvers.isEmpty() ? 0 : exceptionResolvers.size() - 1;
         exceptionResolvers.add(position, new BugsnagMvcExceptionHandler(bugsnag));
     }
--- a/bugsnag-spring/src/main/java/com/bugsnag/MvcConfiguration.java
+++ b/bugsnag-spring/src/main/java/com/bugsnag/MvcConfiguration.java
@@ -1,10 +1,12 @@
 package com.bugsnag;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.HandlerExceptionResolver;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurationSupport;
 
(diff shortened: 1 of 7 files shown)


johnkiely1:
Hi @TobiasMende, 

Thanks for the PR, we will take a look at this as soon as priorities allow.

TobiasMende:
@johnkiely1 Cool! I am looking forward to your feedback. Meanwhile I have added two tests, to verify the behaviour.

bugsnag-spring/src/main/java/com/bugsnag/MvcConfiguration.java
```diff
@@ -1,36 +1,48 @@
...
+     * The first two handlers handle exceptions in an application-specific manner.
+     * (either by @{@link org.springframework.web.bind.annotation.ExceptionHandler}
+     * or @{@link org.springframework.web.bind.annotation.ResponseStatus})
+     * <p>
+     * Therefore, exceptions that are handled by these handlers should not be handled by Bugsnag.
+     * Only unhandled exceptions shall be sent to Bugsnag.
+     */
+    @Override
+    public void extendHandlerExceptionResolvers(List<HandlerExceptionResolver> exceptionResolvers) {
+        final int position = exceptionResolvers.isEmpty() ? 0 : exceptionResolvers.size() - 1;
```
Pezzah, line 45: As far as I can tell the default exception resolvers are populated [here](https://github.com/spring-projects/spring-framework/blob/main/spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/WebMvcConfigurationSupport.java#L1075).

My concern with this approach is that we don't have any control over the contents of this list. If the order or content of the default exception handlers is changed in the spring libraries then it could affect the way that the Bugsnag library works. If it does then we might end up in the situation of trying to support multiple versions of this mechanism as we can't guarantee what versions of the spring framework users are using.

Although there are problems with the current implementation, I believe that it is more resilient to change in the framework, and it's still possible to exclude unwanted exceptions using the `beforeNotify` callback, or by specifying "Discard by error class" in the project settings.

PR #173, opened 2021-12-22
Migrate mazerunner tests to buildkite

fractalwrench:
## Goal

Migrates the mazerunner tests to run on Buildkite. This removes the need for any use to run any tests on Travis CI 🎉 

This alters the fixture to allow deploying the plain Spring app as a .war file to Tomcat.


twometresteve (approved):

.buildkite/pipeline.yml
```diff
@@ -20,3 +20,11 @@ steps:
       - docker-compose#v3.7.0:
           run: java-common
     command: './gradlew check test'
+
+  - label: ':docker: Run mazerunner tests'
+    key: 'java-mazerunner-tests'
+    timeout_in_minutes: 60
+    plugins:
+      - docker-compose#v3.7.0:
+          run: java-mazerunner
```
twometresteve, line 29: As you also seem to be updating to maze v6, I would add a `pull: java-mazeunner` immediately before this - so that it will pick up any new versions of maze runner should be release them while AWS instances are already running.

PR #174, opened 2021-12-23
Update tests to use mazerunner v6

fractalwrench:


features/session_tracking_auto.feature
```diff
@@ -2,25 +2,10 @@ Feature: Auto Session Tracking
...
-    And the payload field "sessionCounts" is an array with 1 element
-    And the payload field "sessionCounts.0.startedAt" is not null
-    And the payload field "sessionCounts.0.sessionsStarted" equals 1
-    And the payload field "app" is not null
-    And the payload field "app.version" equals "1.0.0"
-
-Scenario: Report automatic session from a plain spring app
-    Given I set environment variable "AUTO_CAPTURE_SESSIONS" to "true"
-    And I run the plain spring app
-    When I navigate to the route "/add-session" on port "8080"
```
twometresteve, line 17: Was this removal deliberate?  Is it not needed?

twometresteve (approved):
LGTM - just one question about a removed scenario.

features/session_tracking_auto.feature
fractalwrench: I've removed this scenario as I was struggling to get it working as the way it's written relies on an envar being set to enable automatic session tracking. I also believe the functionality is covered well enough by the previous scenario - generally speaking there's a lot of duplicate tests

Issue #176, opened 2022-05-16
Freeze when initializing Bugsnag


vanniktech:
The SQLDelight IntelliJ Idea plugin uses Bugsnag. When opening the settings from IntelliJ, bug snag gets initialized via sqldelight which blocks the UI.

```
"AWT-EventQueue-0" prio=0 tid=0x0 nid=0x0 runnable
     java.lang.Thread.State: RUNNABLE
 (in native)
	at java.base@11.0.13/java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
	at java.base@11.0.13/java.net.InetAddress$PlatformNameService.lookupAllHostAddr(InetAddress.java:929)
	at java.base@11.0.13/java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1519)
	at java.base@11.0.13/java.net.InetAddress$NameServiceAddresses.get(InetAddress.java:848)
	at java.base@11.0.13/java.net.InetAddress.getAllByName0(InetAddress.java:1509)
	at java.base@11.0.13/java.net.InetAddress.getLocalHost(InetAddress.java:1641)
	at com.bugsnag.callbacks.DeviceCallback.lookupHostname(DeviceCallback.java:44)
	at com.bugsnag.callbacks.DeviceCallback.getHostnameValue(DeviceCallback.java:22)
	at com.bugsnag.callbacks.DeviceCallback.initializeCache(DeviceCallback.java:52)
	at com.bugsnag.Configuration.<init>(Configuration.java:58)
	at com.bugsnag.Bugsnag.<init>(Bugsnag.java:92)
	at app.cash.sqldelight.intellij.SqlDelightErrorHandler.<init>(SqlDelightErrorHandler.kt:32)
...
```

There's also a thread dump:

[threadDumps-freeze-20220512-111722-AI-213.7172.25.2211.8563311-5sec.zip](https://github.com/bugsnag/bugsnag-java/files/8702300/threadDumps-freeze-20220512-111722-AI-213.7172.25.2211.8563311-5sec.zip)

I've reported it to both [Android Studio](https://issuetracker.google.com/issues/232348779) as well as [sqldeilght](https://github.com/cashapp/sqldelight/issues/3230). 

There also seem to be some suggested workarounds:

- https://stackoverflow.com/questions/33289695/inetaddress-getlocalhost-slow-to-run-30-seconds
- https://github.com/gradle/gradle/issues/11122

luke-belton:
[...]

PR #179, opened 2022-06-21
Implement timeout for dns hostname lookup

hamudi-ahmed:
## Goal

This PR aims to fix a freeze reported in #176. 

DNS hostname lookups with `InetAddress.getLocalHost()` can be very slow to run in certain situations and this can cause apps to hang on initialisation when determining `device.hostname`

## Design

Added a 10s timeout when performing the DNS hostname lookup. If it cannot be resolved in that time then the metadata is simply not set. The timeout period here is fairly arbitrary.

## Testing

Couldn't reproduce the slow DNS lookup, but ran unit tests and manually tested with the example app to confirm that DNS hostname still resolves successfully.

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/callbacks/DeviceCallback.java
+++ b/bugsnag/src/main/java/com/bugsnag/callbacks/DeviceCallback.java
@@ -53,7 +53,9 @@ public class DeviceCallback implements Callback {
                 return InetAddress.getLocalHost().getHostName();
             }
         });
-        future.run();
+        Thread resolverThread = new Thread(future, "Hostname Resolver");
+        resolverThread.setDaemon(true);
+        resolverThread.start();
 
         try {
             return future.get(HOSTNAME_LOOKUP_TIMEOUT, TimeUnit.MILLISECONDS);


bugsnag/src/main/java/com/bugsnag/callbacks/DeviceCallback.java
```diff
@@ -39,10 +48,24 @@ private static String lookupHostname() {
...
+        // Resort to dns hostname lookup.
+        // Look up the hostname in a daemon thread, with a timeout of HOSTNAME_LOOKUP_TIMEOUT.
+        // This is to prevent applications hanging waiting for the dns lookup to resolve
+        ExecutorService lookupService = Executors.newSingleThreadExecutor(new DaemonThreadFactory());
+        Future<String> future = lookupService.submit(new Callable<String>() {
+            @Override
+            public String call() throws UnknownHostException {
+                return InetAddress.getLocalHost().getHostName();
+            }
+        });
```
lemnik, line 60:
```suggestion
        FutureTask<String> future = new FutureTask<String>(new Callable<String>() {
            @Override
            public String call() throws UnknownHostException {
                return InetAddress.getLocalHost().getHostName();
            }
        });
        Thread resolverThread = new Thread(future, "Hostname Resolver");
        resolverThread.setDaemon(true);
        resolverThread.start();
```

I'm not convinced that using an `ExecutorService` here is the correct approach, as without shutting it down it will leak both the `ExecutorService` and `Thread`. Using a plain `Thread` gives us the same functionality, but will terminate as soon as its job is complete and in this case it holds no external references so shouldn't cause leaks.

lemnik (changes requested):

bugsnag/src/main/java/com/bugsnag/callbacks/DeviceCallback.java
lemnik, line 60: Instead of adding a timeout to the `lookupHostname` method, maybe we should instead wrap the "startup" cache in a daemon thread?

```java
    public static void initializeCache() {
        Thread hostnameLookup = new Thread("Hostname Lookup") {
            @Override
            public void run() {
                getHostnameValue();
            }
        };
        hostnameLookup.setDaemon(true);
        hostnameLookup.start();
    }
```

lemnik (changes requested):

bugsnag/src/main/java/com/bugsnag/callbacks/DeviceCallback.java
yousif-bugsnag: As discussed, I've moved the cache init to a daemon thread to prevent blocking on initialization, but retained the timeout on the hostname lookup as `getHostnameValue` may also be called from elsewhere before the cache is initialized.

bugsnag/src/main/java/com/bugsnag/callbacks/DeviceCallback.java
```diff
@@ -45,18 +45,15 @@ private static String lookupHostname() {
...
         FutureTask<String> future = new FutureTask<String>(new Callable<String>() {
             @Override
             public String call() throws UnknownHostException {
                 return InetAddress.getLocalHost().getHostName();
             }
         });
-        Thread resolverThread = new Thread(future, "Hostname Resolver");
-        resolverThread.setDaemon(true);
-        resolverThread.start();
+        future.run();
```
lemnik, line 56: This will need to be threaded. `future.run()` will wait for the hostname to be resolved before returning, which means that `future.get(HOSTNAME_LOOKUP_TIMEOUT...` will return immediately (and never timeout).

lemnik (approved):
LGTM

PR #186, opened 2023-03-16
[PIPE-5521] upgrade to use gradle 7.6.1

DavidPetran:
## Goal

This change is necessary to facilitate adding support for Spring version 6/ Spring boot version 3 which require a minimum Gradle version 7.3 and Java 17.

⚠️This is a breaking change as support for java 6 is removed in the bugsnag library, and support for java 7 is removed in the bugsnag-spring library.

## Changeset

- Updated build.gradle files to remove outdated _compile_ dependency configurations.
- Moved to springboot 2.5.14 as compile dependency as it's the minimum version that supports Java 17.
- Replaced javax.annotation.PostConstruct annotation with spring InitializingBean as javax-api can no longer be relied on.
- Moved Mockito forward to fix broken reflection in Java 17 & Fixed deprecation on ArgumentMatchers.anyMapOf
- Always include examples as java 17 must be compatible with JDK 8
- Set :bugsnag source compatibility to 1.7 as 1.6 is no longer supported in Gradle 7
- Set :bugsnag-spring compatibility to 1.8 as minimum version for springboot 2.5.14
- Moved to org.gretty in servlet example as app didn't start with akhikhl.gretty, version 3.1.1 chosen as it supports javax namespace.

## Testing

Existing coverage should suffice to prove nothing has broken. I've also manually tested each example app


bugsnag-spring/src/main/java/com/bugsnag/BugsnagSpringConfiguration.java
```diff
@@ -64,9 +64,8 @@ ScheduledTaskBeanLocator scheduledTaskBeanLocator() {
      * If using Logback, stop any configured appender from creating Bugsnag reports for Spring log
      * messages as they effectively duplicate error reports for unhandled exceptions.
      */
-    @PostConstruct
-    @SuppressWarnings("checkstyle:emptycatchblock")
-    void excludeLoggers() {
+    @Override
+    public void afterPropertiesSet() throws java.lang.Exception {
```
lemnik, line 68: Any specific reason the FQN is being used here?
```suggestion
    public void afterPropertiesSet() throws Exception {
```

lemnik (approved):
LGTM

PR #187, opened 2023-03-23
[PIPE-5522] add jakarta notifier

DavidPetran:
## Goal

This change is necessary to facilitate adding support for Spring version 6/ Spring boot version 3 which require exceptions in the jakarta namespace to be handled.
We want to handle both jakarta and javax namespaces in the same library.

## Design

On the surface it looks like there's some repetition that can be interfaced away. However interfaces that implement both javax and jakarta namespaces would **require** consumer applications to have both (javax and jakarta libraries) on their classpath.
This approach includes support for both libraries but only allows the JVM to load the notifier(s) for the namespace being used by the consuming application.
Separating approaches in this way also increases resiliency for future change to the jakarta library.

## Changeset

- Added jakarta-api to build and renamed javax version name variable 'servletApiVersion'
- Moved javax namespace servlet components to new package com.bugsnag.servlet.javax
- Renamed javax namespace ServletCallback
- Added jakarta namespace servlet components: BugsnagServletContainerInitializer, BugsnagServletRequestListener
- Added jakarta namespace callback
- Added test for jakarta namespace callback
- Added service definition for jakarta namespace initializer
- Renamed javax servlet example application
- Added jakarta servlet example application

## Testing

Manually tested using example apps:
 - servlet-jakarta example app does not include request in payload when using Jetty as requests are destroyed before error pages are called. Consumers should create a custom error handler or filter if they wish to capture this.


lemnik (approved):
LGTM

PR #190, opened 2023-05-26
PIPE-5521: Remove dependency on gradle-nexus-plugin

DavidPetran:
## Goal

Enable the command `./gradlew -Preleasing=true publishToMavenLocal` as this was missed when initially upgraded to gradle v7. 

## Design

Similar approach to BAGP

## Changeset

Removed the gradle-nexus-plugin and replaced with pom directive.
Prevent release of subprojects spring:javax and spring:jakarta using project.depth
Ensure spring:javax jar exists before packaging bugsnag-spring using 'dependsOn' 

## Testing

Tested under PLAT-9981

### DIFF
--- a/bugsnag-spring/build.gradle
+++ b/bugsnag-spring/build.gradle
@@ -33,7 +33,7 @@ dependencies {
 
 evaluationDependsOnChildren()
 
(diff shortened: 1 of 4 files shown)


lemnik (approved):
LGTM

PR #193, opened 2023-06-06
Release v3.7.0

yousif-bugsnag:


tomlongridge (approved):

pc-bob:
It looks like this did not actually generate a release, was that intentional? Any idea when this new version will be released?

yousif-bugsnag:
Hi @pc-bob, 

We had some technical issues getting the build published - we are now planning to release this early next week if you want to check back then.

PR #197, opened 2023-06-27
Fix minimum JVM requirement

lemnik:
## Goal
Enable users of Java 8 to build their projects using Gradle (which checks the required JVM version by using the Gradle `.module` file).

> **Note**
> Don't be intimidated by the number of files in this PR. Most of the changes are simply moving files to arrange the sources into the new  sourceSet structure.

## Design
Merged the `bugsnag-spring` and `bugsnag-spring:javax` projects back together in favor of using a split `sourceSet` system with Gradle managed tool chains. This allows us to define a structure similar to that used by Kotlin-Multiplatform:

- `src/common` - common to both the older (`javax`) and new `jakarta` implementations of the `bugsnag-spring` notifier, built using a Java 8 toolchain
- `src/javax` - the implementation classes that compile against the `javax` packages, depending on `common` and built using a Java 8 toolchain
- `src/jakarta` - implementation classes that compile against the `jakarta` packages and are compatible with Spring Boot 3, built using a Java 17 toolchain

A Gradle `configuration` is defined for each of these `sourceSets` in order to avoid the default behaviors of Gradles intrinsic Java `sourceSet` configurations and give us finer control over the exact nature of our classpaths (where both `javax` and `jakarta` include `common`).

## Changeset

- Merged the `bugsnag-spring:javax` module back into `bugsnag-spring` as a sourceSet and configuration
- Split the `bugsnag-spring` project into 3 sourceSets: common, javax, jakarta
- Split the unit tests for `bugsnag-spring` into `javaxTest` and `jakartaTest`
- Used Gradle toolchains to ensure that `bugsnag-spring` builds under Java 8, with the `jakarta` related sources building under Java 17
- Changed the `versionMapping` of the publication to use the version resolved from the `javax` configuration (instead of the latest used by the entire project)

## Testing
The existing tests pass (although I had to update our standard `logback.xml` configuration, which included an option that was removed a while back).

Manually tested publishing locally to check the module version metadata was correct:
image


bugsnag-spring/build.gradle
```diff
@@ -1,52 +1,186 @@
...
-    testImplementation project(':bugsnag')
-    testImplementation "junit:junit:${junitVersion}"
-    testImplementation "jakarta.servlet:jakarta.servlet-api:${jakartaServletApiVersion}"
-    testImplementation "org.springframework.boot:spring-boot-starter-test:${springBootVersion}"
-    testImplementation "org.springframework.boot:spring-boot-starter-web:${springBootVersion}"
-    testImplementation "org.mockito:mockito-core:${mockitoVersion}"
+sourceSets {
+    common {
+        java
+        compileClasspath += configurations.compileCommon + configurations.compileJavax
```
tomlongridge, line 32: Is it correct that the common sourceSet has `compileJavax`?

tomlongridge (commented):
Looks good from my perspective - just one query - seems a much neater structure.

bugsnag-spring/build.gradle
lemnik: Yes. This allows the common classes to access the Spring classes that are not specific to either version. Unfortunately there isn't really a cleaner way to do this without complicating the dependencies of common and confusing things further.

tomlongridge: OK - suggest we add a comment here as a reminder to future readers.

tomlongridge (approved):

PR #215, opened 2024-06-18
Replace filters with redactedKeys

clr182:
## Goal

Notifier spec has changed to favour redactedKeys over filters. Change is in place to keep in line with notifier spec. Additionally this changes the behaviour from case sensitive Strings to a  pattern matching regex implementation.


clr182 (commented):
Re-add Filters

PR #217, opened 2024-07-01
Adding Redacted Keys

clr182:
## Goal

Filtering config option is case sensitive and is no longer in line with our notifier spec; as such it should be deprecated in favour of `redactedKeys`.

## Design

Added `redactedKeys` configuration option, which is case insensitive, as a replacement for the `filters` configuration option. Additionally due to conflictions the default `filters` parameters were changed.

## Changeset

* Added `redactedKeys` config option.

* Made `redactedKeys` case insensitive.

* Changed `filters` config default parameters from 
``` {"password", "secret", "Authorization", "Cookie"};``` 
to 
```{"ipAddress", "logLevel"};```

* Set `refactedKeys` config default parameters to
``` {"password", "secret", "Authorization", "Cookie"};``` 

## Testing

Added and updated tests for filtering mechanism and redactedKeys mechanism.

### CODE REVIEW

**Comment by clr182** [comment]
May have to discuss this change with reviewer. It was causing conflicts in tests as these two methods are so similar.

```
Changed filters config default parameters from
 {"password", "secret", "Authorization", "Cookie"};
to
{"ipAddress", "logLevel"};
```

File: bugsnag/src/test/resources/logback.xml [hunk 1]
```diff
@@ -7,8 +7,8 @@
 
         <appType>gradleTask</appType>
 
-        <filteredProperty>password</filteredProperty>
-        <filteredProperty>credit_card_number</filteredProperty>
+        <redactedKeys>password</redactedKeys>
```
tomlongridge, line 10: I think this should probably be singular in this context? Like `ignoredClass`?

```suggestion
        <redactedKey>password</redactedKey>
```

bugsnag/src/test/java/com/bugsnag/AppenderTest.java
```diff
@@ -144,10 +144,10 @@ public void testBugsnagConfig() {
...
         assertFalse(config.shouldAutoCaptureSessions());
 
-        assertEquals(2, config.filters.length);
-        ArrayList<String> filters = new ArrayList<String>(Arrays.asList(config.filters));
-        assertTrue(filters.contains("password"));
-        assertTrue(filters.contains("credit_card_number"));
+        assertEquals(2, config.redactedKeys.length);
+        ArrayList<String> redactedKeys = new ArrayList<String>(Arrays.asList(config.redactedKeys));
+        assertTrue(redactedKeys.contains("password"));
+        assertTrue(redactedKeys.contains("credit_card_number"));
```
tomlongridge, line 150: It would be worth adding a few tests to verify the expected interplay between `filters` and `redactedKeys`

bugsnag/src/test/java/com/bugsnag/BugsnagTest.java
```diff
@@ -164,8 +164,8 @@ public void close() {
     }
 
     @Test
-    public void testFilters() {
-        bugsnag.setFilters("testfilter1", "testfilter2");
+    public void testRedactedKeys() {
```
tomlongridge, line 167: We should keep the filters tests as well to ensure that doesn't break whilst we still have both.

features/redactedKeys.feature
tomlongridge, line 1: Suggest keeping the `filering_metadata.feature` as well, whilst the functionality is still there.

bugsnag/src/test/java/com/bugsnag/AppenderTest.java
clr182: I've added functionality so that when both are called, setRedactedKeys creates a temp array to append the filters to the redactedKeys config option to avoid overwriting.

bugsnag/src/main/java/com/bugsnag/Bugsnag.java
```diff
@@ -232,14 +232,32 @@ public void setEndpoint(String endpoint) {
...
+     * Set which keys should be redacted when sending metaData to Bugsnag.
+     * Use this when you want to ensure sensitive information, such as passwords
+     * or credit card information is stripped from metaData you send to Bugsnag.
+     * Any keys in metaData which contain these strings will be marked as
+     * [REDACTED] when send to Bugsnag.
+     *
+     * @param redactedKeys a list of String keys to be redacted from metaData
+     */
+    public void setRedactedKeys(String... redactedKeys) {
+        config.redactedKeys = redactedKeys;
```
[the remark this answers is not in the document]
tomlongridge: Looking at the changes, I wonder if in fact this change is more risky as a non-breaking change than I thought as we'll need to compile the filter strings as regexes (L249). At very least we need a try-catch for a `PatternSyntaxException`, but it makes me a bit nervous even if 99.9% of users will have pretty straightforward key names with an occasional `.` which will still match a literal `.`.

@lemnik - thoughts here? The alternative is that we go back to running the two config options independently alongside each other (which I think was @clr182's first take).

PR #218, opened 2024-07-02
add aopproxyutils dependancy and add handling in scheduledTaskConfig class

clr182:
## Goal

The following warning appears when starting the SpringBootWeb application.
“com.bugsnag.ScheduledTaskConfiguration : Bugsnag scheduled task exception handler could not be configured”.
This happens when adding the `@EnableScheduling` annotation

## Design

Checked the ScheduledTaskRegister to determine whether it is a proxy of the original object, then get the target class of the original object if it is and cast it as the taskSchedular before configuring the existing taskScheduler.

## Changeset

Added new AoP springboot dependencies for accessing the original ScheduledTaskSchedular object.

Added checking to the ConfigureTasks method in ScheduledTaskConfiguration.java to determine whether the passed object is a proxy or not. 

Handle cases where this is or isn't a proxy within the configureTasks method

## Testing

Added two tests to determine if the taskScheduler object is a proxy.

### DIFF
--- a/bugsnag-spring/src/jakartaTest/java/com/bugsnag/ScheduledTaskConfigurationTest.java
+++ b/bugsnag-spring/src/jakartaTest/java/com/bugsnag/ScheduledTaskConfigurationTest.java
@@ -3,6 +3,7 @@ package com.bugsnag;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.mockito.Mockito.when;
 
 import com.bugsnag.testapp.springboot.TestSpringBootApplication;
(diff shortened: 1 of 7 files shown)


bugsnag-spring/src/jakartaTest/java/com/bugsnag/ScheduledTaskConfigurationTest.java
```diff
@@ -113,6 +115,31 @@ public void findExecutorByName() throws NoSuchFieldException, IllegalAccessExcep
...
+                "errorHandler should be BugsnagScheduledTaskExceptionHandler",
+                errorHandler instanceof BugsnagScheduledTaskExceptionHandler
+        );
+    }
+
+    private TaskScheduler createProxy(TaskScheduler target) {
+        ProxyFactory factory = new ProxyFactory(target);
+        factory.addAdvice((MethodInterceptor) invocation -> invocation.proceed());
+        return (TaskScheduler) factory.getProxy();
+    }
```
lemnik, line 141: This common method can probably be extracted to a util in `commonTest`

lemnik (approved):
LGTM

PR #219, opened 2024-07-14
Update Serializer object to be an interface

clr182:
## Goal

Sometimes a non-serializable object can be passed as metadata, which can cause an error. As such the Serializer class should be converted to an interface so a user may use their own serializer which inherits from the Serializer interface. 

## Design

To allow users to develop their own Serializer for handling non-serializable data.

## Changeset

Created an ISerializer class and changed the Serializer class to implement it. 
By default this creates a default serializer if it is not specified in the configuration option.

## Testing

Added test `testCustomSerializer` to `ConfigurationTest.java` which creates an anonymous class extending the `DefaultSerializer` and overwrites the `WriteToStream` method to write `foo`. Added an `assertTrue` call if the method has been called and added a call to`assertEqual` to confirm that the expected value is the same as the output.


bugsnag/src/main/java/com/bugsnag/MetaData.java
```diff
@@ -1,12 +1,22 @@
...
 import java.util.Map;
+import java.util.logging.Logger;
 
 class MetaData extends HashMap<String, Object> {
     private static final long serialVersionUID = 2530038179702722770L;
+    private static Logger logger = Logger.getLogger(MetaData.class.getName());
 
     public void addToTab(String tabName, String key, Object value) {
+        //check if the metadata to add is serializable
+        if (!(value instanceof Serializable)) {
```
tomlongridge, line 14: As discussed IRL, this doesn't mean the same as being serializable for Jackson to JSON. Let's take an alternative approach and make the serialiser in configuration public and docs/tests for customizing it to serialise a custom object.

bugsnag/src/main/java/com/bugsnag/serialization/ISerializer.java
```diff
@@ -0,0 +1,23 @@
+package com.bugsnag.serialization;
+
+import java.io.OutputStream;
+
+public interface ISerializer {
```
tomlongridge, line 5: We should be fine to call this `Serializer` and rename the existing implementation as `DefaultSerializer` - although it was a public class, there was no way to use it so we can replace the class name with the interface without issue.

bugsnag/src/main/java/com/bugsnag/serialization/ISerializer.java
```diff
@@ -0,0 +1,23 @@
...
+    void writeToStream(OutputStream stream, Object object) throws SerializationException;
+
+    /**
+     * Convert the object to a JSON string.
+     *
+     * @param object the object to convert to JSON.
+     * @return the JSON string representation of the object.
+     * @throws SerializationException the object could not be serialized.
+     */
+    String toJson(Object object) throws SerializationException;
```
tomlongridge, line 22: Is this only required for test? If so, I don't think we'd want to include in the interface – we can either test it against the concrete implementation or make the test use the output stream.

bugsnag/src/main/java/com/bugsnag/Configuration.java
```diff
@@ -46,12 +47,17 @@ public class Configuration {
...
-    Serializer serializer = new Serializer();
+    private ISerializer serializer;
     private final AtomicBoolean autoCaptureSessions = new AtomicBoolean(true);
     private final AtomicBoolean sendUncaughtExceptions = new AtomicBoolean(true);
 
     Configuration(String apiKey) {
+        this(apiKey, new Serializer());
+    }
+
+    Configuration(String apiKey, ISerializer serializer) {
```
tomlongridge, line 58: We only have API key as a parameter for configuration - all the others should be initialised with a default (i.e. the default Serializer implementation) and be set via properties.

bugsnag/src/test/java/com/bugsnag/ConfigurationTest.java
```diff
@@ -95,6 +103,27 @@ public void testInvalidSessionEndpoint() {
         assertEquals("http://sessions.example.com", getDeliveryEndpoint(config.sessionDelivery));
     }
 
+    @Test
+    public void testSerializeObject() throws SerializationException {
```
tomlongridge, line 107: This doesn't feel like the right level of test - the unit tests are testing that configuration is set right, rather than that what is provided does the right thing. So you could pass your own implementation of `Serializer` in that simply writes "foo" to the output stream and then validate that this is what you receive back; but we shouldn't compare JSON outputs.

bugsnag/src/main/java/com/bugsnag/Configuration.java
```diff
@@ -183,4 +189,20 @@ Map<String, String> getSessionApiHeaders() {
...
+        this.serializer = serializer;
+    }
+
+    /**
+     * Get the serializer used for serializing custom objects.
+     * @return the serializer
+     */
+    public ISerializer getSerializer() {
+        return serializer;
+    }
```
tomlongridge, line 207: It seems that simple getting/setting on the Configuration class is done directly using public fields rather than providing methods - the existing setter functions are for more complicated logic, but this only requires assignement.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 * Add a null check for `Severity` to the notify override method. [#214](https://github.com/bugsnag/bugsnag-java/pull/214)
 
+* Add handling for non serializable object passed to metadata. [#219](https://github.com/bugsnag/bugsnag-java/pull/219)
```
tomlongridge, line 11: This and the PR description need updating for the amended approach - we're not handling non-serializable objects directly: just providing a way for people to do so if they provide their own serializer.

bugsnag/src/test/java/com/bugsnag/SerializerTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.junit.Test;
+
+import java.io.ByteArrayOutputStream;
+
+public class SerializerTest {
+
+    @Test
+    public void testCustomSerializer() throws SerializationException {
+        Serializer customSerializer = new CustomSerializer();
```
tomlongridge, line 17: You're not really testing anything here - the class you're calling is a test implementation, so it's just asserting that the compiler can cope with interfaces :) 

Instead, I'd put this test in with the Configuration unit tests and have your `writeToStream` implementation (suggest using an anonymous class for readability of the test) assert that it's called.

features/meta_data.feature
tomlongridge, line 1: Revert changes to avoid touching the file.

bugsnag/src/main/java/com/bugsnag/serialization/Serializer.java
tomlongridge, line 1: It would be worth Javadoc on this class, now that it's public.

> The Serializer is called to generate the JSON for an object to be added as metadata to a Bugsnag event. 

... and a variant of this for the `writeToStream` method.

bugsnag/src/main/java/com/bugsnag/Configuration.java
```diff
@@ -46,13 +47,13 @@ public class Configuration {
...
     Collection<Callback> callbacks = new ConcurrentLinkedQueue<Callback>();
-    Serializer serializer = new Serializer();
+    Serializer serializer;
     private final AtomicBoolean autoCaptureSessions = new AtomicBoolean(true);
     private final AtomicBoolean sendUncaughtExceptions = new AtomicBoolean(true);
 
     Configuration(String apiKey) {
         this.apiKey = apiKey;
-
+        this.serializer = new DefaultSerializer();
```
tomlongridge, line 56: I can't see any reason why this can't be a public property (defined a little higher-up) and assigned to on the same line as being declared (so it fits the existing style).

Also, we don't need a setter (at least I don't think we do)

CHANGELOG.md
tomlongridge: The PR title and description are still out of date

bugsnag/src/main/java/com/bugsnag/Configuration.java
```diff
@@ -46,13 +47,12 @@ public class Configuration {
     public boolean sendThreads = false;
```
tomlongridge, line 47: I meant to move it to the above list so that it's declared in the same way as the other config options.

```suggestion
    public boolean sendThreads = false;
    public Serializer serializer = new DefaultSerializer();
```

tomlongridge (approved):

PR #227, opened 2025-06-26
Set default endpoints based on API key

richardelms:
## Goal

Set default endpoints based on API key. If no endpoint is configured, payloads should be sent to `*.insighthub.smartbear.com` if the API key starts with `00000`.

## Testing

Aded Unit Tests


bugsnag/src/main/java/com/bugsnag/delivery/SyncHttpDelivery.java
```diff
@@ -52,6 +55,32 @@ public void setTimeout(int timeout) {
...
+     *
+     * @param key 32-character Bugsnag API key
+     * @return Insight Hub sessions endpoint when the key starts with “00000”,
+     *     otherwise the classic Bugsnag endpoint.
+     */
+    public static String defaultSessionFor(String key) {
+        return (key != null && key.startsWith(HUB_KEY_PREFIX))
+                ? HUB_SESSION_ENDPOINT
+                : DEFAULT_SESSION_ENDPOINT;
+    }
```
lemnik: I think these methods would be better added to the `HttpDelivery` interface rather than this specific implementation.

I also think that we should probably name them in a more "Java" pattern? Something like: `resolveDefaultSessionEndpoint`

bugsnag/src/main/java/com/bugsnag/Configuration.java
```diff
@@ -120,23 +123,36 @@ public boolean shouldSendUncaughtExceptions() {
...
-     * https://notify.bugsnag.com, and sessions to https://sessions.bugsnag.com, but you can
-     * override this if you are using Bugsnag Enterprise to point to your own Bugsnag endpoint.
-     *
+     * @deprecated Use {@link #setEndpoints(EndpointConfiguration)} instead.
+     */
+    @Deprecated
+    public void setEndpoints(String notify, String sessions) throws IllegalArgumentException {
+        if (notify == null || notify.isEmpty() || sessions == null || sessions.isEmpty()) {
+            throw new IllegalArgumentException("endpoints cannot be empty or null.");
+        }
```
lemnik: It bothers me that this check doesn't line-up with the `EndpointConfiguration` constructor. Given that the constructor does the null checks again, is it possibly worth skipping this block of checks completely?

richardelms: agreed 👍

lemnik (approved):

PR #241, opened 2025-11-14
rename callbacks and update functionality

richardelms:
Rename `beforeNotify` -> `onError` and `beforeSendSession` -> `onSession`
Make callbacks return a bool so that sessions and events can be vetoed

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/OnSession.java
+++ b/bugsnag/src/main/java/com/bugsnag/OnSession.java
@@ -2,5 +2,5 @@ package com.bugsnag;
 
 @FunctionalInterface
 interface OnSession {
-    Boolean onSession(SessionPayload payload);
+    boolean onSession(SessionPayload payload);
 }
--- a/bugsnag/src/main/java/com/bugsnag/OnSession.java
+++ b/bugsnag/src/main/java/com/bugsnag/OnSession.java
@@ -1,5 +1,5 @@
 package com.bugsnag;
 
-interface BeforeSendSession {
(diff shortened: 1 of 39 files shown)


bugsnag/src/main/java/com/bugsnag/OnSession.java
```diff
@@ -0,0 +1,5 @@
+package com.bugsnag;
+
+interface OnSession {
```
lemnik, line 4: I think it might be worth annotating this with `@FunctionalInterface` as a "SAM" guarantee:

```suggestion
@FunctionalInterface
interface OnSession {
```

bugsnag/src/main/java/com/bugsnag/OnSession.java
```diff
@@ -0,0 +1,5 @@
...
+    void onSession(SessionPayload payload);
```
lemnik: On other platforms `onSession` returns a `boolean` to allow sessions to be suppressed

```suggestion
    boolean onSession(SessionPayload payload);
```

lemnik (commented):
I think it's worth going over the test callbacks and replacing the old anonymous inner class syntax `new Callback() {` with lambdas. Maybe in a future PR?

bugsnag/src/main/java/com/bugsnag/OnSession.java
richardelms: this has been implemented for session and error callbacks now

lemnik (approved):
LGTM

PR #242, opened 2025-11-24
Rename endpointConfiguration to endpoints

richardelms:
Rename endpointConfiguration to endpoints

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/Configuration.java
+++ b/bugsnag/src/main/java/com/bugsnag/Configuration.java
@@ -37,7 +37,7 @@ public class Configuration {
     public Delivery delivery;
     public EndpointConfiguration endpoints;
     public Delivery sessionDelivery;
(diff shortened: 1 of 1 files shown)


bugsnag/src/main/java/com/bugsnag/Configuration.java
```diff
@@ -35,7 +35,7 @@ public class Configuration {
     public String appType;
     public String appVersion;
     public Delivery delivery;
-    public EndpointConfiguration endpointConfiguration;
+    public EndpointConfiguration endpoints;
```
lemnik, line 38: Given that we already have `setEndpoints` methods in the `Configuration` class, maybe we should also make the field `private` (we should probably add a `getEndpoints()` as well in that case)?

```suggestion
    private EndpointConfiguration endpoints;
```

richardelms: i'm just about to do the encapsulation task and will make sure to get these accessors right there 👍

PR #243, opened 2025-11-24
Encapsulation

richardelms:
Improve encapsulation and maintainability by replacing direct field access in the `Configuration` class with getter and setter methods. 

Also update metadata accessors to be closer to spec compliance.

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/Bugsnag.java
+++ b/bugsnag/src/main/java/com/bugsnag/Bugsnag.java
@@ -220,8 +220,9 @@ public class Bugsnag implements Closeable {
      */
     @Deprecated
     public void setEndpoint(String endpoint) {
-        if (config.getDelivery() instanceof HttpDelivery) {
-            ((HttpDelivery) config.getDelivery()).setEndpoint(endpoint);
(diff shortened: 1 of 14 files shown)


bugsnag/src/main/java/com/bugsnag/Bugsnag.java
```diff
@@ -556,8 +553,8 @@ public boolean shouldAutoCaptureSessions() {
      */
     @Deprecated
     public void setSessionEndpoint(String endpoint) {
-        if (config.sessionDelivery instanceof HttpDelivery) {
-            ((HttpDelivery) config.sessionDelivery).setEndpoint(endpoint);
+        if (config.getSessionDelivery() instanceof HttpDelivery) {
+            ((HttpDelivery) config.getSessionDelivery()).setEndpoint(endpoint);
```
lemnik: nit: it's probably a good idea to capture these to avoid calling twice

```suggestion
        Delivery delivery = config.getSessionDelivery()
        if (delivery instanceof HttpDelivery) {
            ((HttpDelivery) delivery).setEndpoint(endpoint);
```

lemnik (approved):
LGTM

PR #251, opened 2026-03-25
Rename sendUncaughtExceptions -> autoDetectErrors

lemnik:
## Goal
Rename sendUncaughtExceptions -> autoDetectErrors

## Testing
Relied on existing tests

### DIFF
--- a/bugsnag/src/main/java/com/bugsnag/ExceptionHandler.java
+++ b/bugsnag/src/main/java/com/bugsnag/ExceptionHandler.java
@@ -1,46 +1,56 @@
 package com.bugsnag;
 
 import java.lang.Thread.UncaughtExceptionHandler;
-import java.util.Set;
-import java.util.WeakHashMap;
+import java.lang.ref.WeakReference;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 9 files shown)


bugsnag/src/main/java/com/bugsnag/ExceptionHandler.java
```diff
@@ -1,46 +1,56 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Stream;
 
 
 class ExceptionHandler implements UncaughtExceptionHandler {
+    private static volatile ExceptionHandler singletonInstance = null;
+
     private final UncaughtExceptionHandler originalHandler;
-    private final WeakHashMap<Bugsnag, Boolean> clientMap = new WeakHashMap<Bugsnag, Boolean>();
+    private final AtomicReference<WeakReference<Bugsnag>[]> weakClients = new AtomicReference<>(null);
```
SB-sachinp, line 17: AtomicReference used here for heavy read or for lock free read?

lemnik: `AtomicRef` is used here for a CAS loop in `add` / `remove`, reads massively outweigh writes and the general volumes should be low. Keeping the `Bugsnag[]` effectively immutable also avoids the need to synchronize during iteration.

SB-sachinp (approved):
Looks good to me.

PR #256, opened 2026-04-28
Removed the duplicate apply(release.gradle.kts)

lemnik:
## Goal
Unblock release by fixing the Gradle script.

## Design
Remove the duplicate `apply("release.gradle.kts")` from the `common.gradle.kts` file. Each module needs to now include `release.gradle.kts` manually (which they were already doing).

## Testing
Published in release mode locally.

### DIFF
--- a/common.gradle.kts
+++ b/common.gradle.kts
@@ -1,9 +1,6 @@
 // Version catalog is now used - see gradle/libs.versions.toml
 apply(plugin = "checkstyle")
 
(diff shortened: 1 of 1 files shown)


SB-sachinp (approved):
Done

SB-sachinp (approved):
Approved

SB-sachinp (approved):
Finally done.

